PR #9, opened 2014-06-30
Added simple support for table-wide data attributes

greenbender:
I haven't added a link showing the feature, however, there is a brief discussion of the issue here http://datatables.net/forums/discussion/17320/setting-options-in-html-attributes-possible

This is a patch I have used myself and is more of a 'proof of concept' than a finished product. There are probably a lot of things I have missed.


DataTables:
Finally done - thanks for your input on this both, particularly your suggested patches @greenbender. I've done it slightly differently:

1. The main init options extend without using a new object to try and avoid breaking any references
2. The columns are extended in the column options functions just to save a bit of code

It should work identically to your own suggestions though.

Thanks again!

PR #11, opened 2014-07-22
Ability to use a function as a selector in the API

kasthor:
This is a quick and easy update, may be you want to fix a few things, I've tested it with my actual use case in which i use the data to attached to the rows, for which It's then an settings.aoData[x]._aData ( I'm not sure if in other cases that's true

Ok, for the use case, its quite simple, right now you have the ability to filter the rows by jQuery selectors, or by number.. but with this patch you would be able to do something like 

```javascript
_dt.api().row( function( idx, node, data){
  return data.get("id") == 150;
}).....
```
And that would get you the row representing the data model with id 150

In my usage case I'm using Backbone, and I'm attaching the model to the row, for this reason I can just call the backbone function _.get("id")_ but this should actually send whatever is attached to that particular data row

Hope you like this idea

Cheers!

Giancarlo

### DIFF
--- a/js/api/api.rows.js
+++ b/js/api/api.rows.js
@@ -49,6 +49,15 @@ var __row_selector = function ( settings, selector, opts )
 			}
 		}
 
+		// if the selector is a function then it wraps it into a new function that will then attach the data object to it
+
+		if ( typeof sel === 'function' ) {
+			var _sel = sel;
+			sel = function( idx, node ){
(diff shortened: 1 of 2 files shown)


DataTables:
Apologies - I added this ability back in 1.10.3 and completely forgot to update this issue! Details about using the row selector as a function is available here: http://datatables.net/reference/type/row-selector#Function .

DataTables:
Not sure why I didn't close this with my last comment... doing so now :-)

PR #16, opened 2014-08-27
Mentioning that ajax.url() accepts function

schakko:
The API documentation misses that ajax.url() accepts although a function like options.ajax

### DIFF
--- a/docs/api/ajax.url().xml
+++ b/docs/api/ajax.url().xml
@@ -22,7 +22,51 @@
 			DataTables.Api instance
 		</returns>
 	</type>
+	
+	<type type="function">
+		<signature>ajax.url(function(data, callback, settings) {})</signature>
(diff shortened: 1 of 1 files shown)


DataTables:
Its taken me a while, but I've finally gotten around to reviewing this! I'm inclined not to pull it in in its current form as `ajax.url()` implies that you are setting only the URL, but actually this provides the option to configure the Ajax option totally, so it should really just be `ajax()`.

I'm not adverse to the basic principle of being able to configure the `ajax` option fully on-the-fly but I think `ajax.url()` is probably the wrong place to do it.

I've added it to my to-do list to take a look into adding an `ajax()` method to DataTables and Editor that will be fully supported for this kind of action. I suspect they should be relatively simple to do... (but I always think that!)

PR #20, opened 2014-10-30
BugFix: Prefer ajax over sAjaxSource

knabar:
According to a comment in https://github.com/DataTables/DataTablesSrc/blob/master/js/api/api.ajax.js#L121, `ajax` has priority over `sAjaxSource`.

This is not true in https://github.com/DataTables/DataTablesSrc/blob/master/js/core/core.ajax.js#L103 if `fnServerData` is set, where only `sAjaxSource` is used.

I ran into this bug in combination with https://github.com/l-lin/angular-datatables, where even after changing the `sAjaxSource` property the initial url is used in all subsequent calls to `fnServerData`.

### DIFF
--- a/js/core/core.ajax.js
+++ b/js/core/core.ajax.js
@@ -100,7 +100,7 @@ function _fnBuildAjax( oSettings, data, fn )
 	{
 		// DataTables 1.9- compatibility
 		oSettings.fnServerData.call( instance,
-			oSettings.sAjaxSource,
+			ajax || oSettings.sAjaxSource,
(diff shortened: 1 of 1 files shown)


DataTables:
Going to close this issue as I think mixing the old and new style options could get quite confusing. The `ajax` option in DataTables can be used for everything the old `fnServerData` option can be.

knabar:
Thanks for considering this - I ended up not using the third-party library that is triggering this bug (https://github.com/l-lin/angular-datatables).  I'll see if it makes sense to report it there.

PR #30, opened 2015-01-27
TypeError: s.match is not a function

neoglez:
When initializing DataTables with 
```javascript
"ajax": {
	        "url": "/some/action/index",
	        "type": "POST"
	    },
"paging": true,
"dom": 'T<"clear">lrtip',
"serverSide": true,
"processing": true,
"scrollX": true,
```
Actually it seem's like it doesn't matter the value of "scrollX" at all, if you include it as part of the options the type error appears. I don't know if the problem here is of a deeper nature but this type test should always be present anyway and in my case it fixed the problem.

### DIFF
--- a/js/core/core.sizing.js
+++ b/js/core/core.sizing.js
@@ -317,6 +317,10 @@ function _fnStringToCss( s )
 			'0px' :
 			s+'px';
 	}
+	
+	if (typeof s.match == 'undefined') {
+		return null;
+	}
 
 	// Check it has a unit character already
 	return s.match(/\d$/) ?


DataTables:
Seems to work as expected for me: http://live.datatables.net/zaxudafe/1/edit . Can you link to a test case showing the issue please?

neoglez:
It seems that it happens when some width is exceeded. I don't know the structure of the response of the script, that's why i just repeat it the column.
[http://live.datatables.net/suloqazi/1/edit]
By the way, the PR avoid a total blocking but it doesn't actually solve the problem :( becouse the data is being scrolled but the headers dont.

DataTables:
Its more that you have defined 7 columns in the `columns` array but there are only 6 in the HTML. They must match otherwise DataTables doesn't know what is going on! This isn't anything to do with scrollX, but rather just a configuration error. I do need to look at improving the error handling to report this issue back, but I'm going to close the PR now as that isn't the issue.

neoglez:
I managed to found the problem and i am going to post here in case of someone having the same problem. jQuery UI (version 1.8.21) was interfering with outerWidth() [making that function return an object] (http://bugs.jquery.com/ticket/12491), so update your JQuery UI and everything should be fine.

DataTables:
Ah yes - that was a good one that. They used a private jQuery method that wasn't meant for internal use and this issue was the result. Thanks for spotting the connection with the old version of jQuery UI and adding the note here!

PR #32, opened 2015-02-02
Added images to bower 'main'section

abrom:
I did some searching through the forums and issues but couldn't find any reason why the images shouldn't be included in the Bower `main` section. This simply adds that folder (the PNGs) to the list of files that should be included in a production install.


DataTables:
Thanks!

PR #35, opened 2015-02-19
Call _fnAjaxDataSrc before setting recordsTotal

nathanstitt:
_fnAjaxDataSrc will in turn call _fnGetObjectDataFn, which
can be specified by the dataSrc method.

This way the dataSrc can set the recordsTotal parameter on
the json data and the changes will be read into the
recordsTotal/recordsFiltered variables.

I've tested this on my development site and it works fine there.

In my case the server does not return the proper "recordsTotal" method
but it can be calculated by other means.

If you'd prefer an alternative method, perhaps the recordsTotal could be
extended to optionally call a function in order to retrieve the counts.  I'll
take a stab at that if you think it'd be better.

### DIFF
--- a/js/core/core.ajax.js
+++ b/js/core/core.ajax.js
@@ -1,5 +1,3 @@
-
-
 /**
  * Create an Ajax call based on the table's settings, taking into account that
  * parameters can have multiple forms, and backwards compatibility.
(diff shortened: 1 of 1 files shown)


DataTables:
Sounds like a good idea to me - thanks for the PR!

Can you just confirm that you are happy for your change to be released under the MIT license? (I really need to get around to seeing up a CLA...)

nathanstitt:
Yes, MIT is great.  Thanks!

PR #37, opened 2015-02-26
hotfix: row attributes via server-side processing

sjoorm:
Since `<tr>` element was existing only in memory and not in DOM tree at the moment of calling `_fnRowAttributes()` using `$(tr)` caused creation of a _copy_ of `<tr>` element and making data-attributes insertion in new `$(tr)` made no sense.
Fixed to use native JS methods to make necessary additions to the target row element.

### DIFF
--- a/js/core/core.draw.js
+++ b/js/core/core.draw.js
@@ -91,28 +91,30 @@ function _fnRowAttributes( row )
 	var data = row._aData;
 
 	if ( tr ) {
-		if ( data.DT_RowId ) {
-			tr.id = data.DT_RowId;
-		}
+		var tr = row.nTr;
+		var data = row._aData;
+	
+		if ( tr ) {
+			if ( data.DT_RowId ) {
+				tr.id = data.DT_RowId;
+			}
+	
+			if ( data.DT_RowClass ) {
+				// Remove any classes added by DT_RowClass before
+				var a = data.DT_RowClass.split(' ');
+				row.__rowc = row.__rowc ?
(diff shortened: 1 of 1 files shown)


DataTables:
Hi - thanks for this. I'll try to review it early next week.

sjoorm:
@DataTables did you have any chance to check this PR?

DataTables:
Not yet I'm afraid. Overwhelmed with support requests and barely able to keep on top of development to be honest!

I'll try to make time later today.

In the mean time, if I do pull this in are you happy for it to be under the current MIT license?

DataTables:
A couple of points:

1) You've used `classList`, which is a fantastic addition to the DOM, but DataTables needs to support IE6+ currently (v1.11 will be IE8+). As such `classList` is not appropriate in DataTables core.

2) The functionality has been changed from using jQuery `$().data()` to writing attributes.

3) I don't understand this point:

> Since `<tr>` element was existing only in memory and not in DOM tree at the moment of calling `_fnRowAttributes()` using `$(tr)` caused creation of a _copy_ of `<tr>` element and making data-attributes insertion in new `$(tr)` made no sense.

It doesn't matter if it is in the DOM tree or not, it is a valid node that will be inserted into the document in future. `$()` will not create a clone of a node unless you tell it to by passing in a string representation or using `$().clone()`. For example:

```
var node = document.createElement('div');
var result = node === $(node)[0];
```

`result` in the above is true.

I won't be pulling this request in because of 1 and 2, but I would like to understand your intention with 3 if you could explain it a bit further please?

PR #38, opened 2015-02-27
Add brazilian currency support for sorting

rodrigoberriel:
The [brazilian currency's (real) symbol](http://en.wikipedia.org/wiki/Brazilian_real) is "R$". As only "$" is escaped, in this case, based on [columns.type](https://datatables.net/reference/option/columns.type), any column that has a R$ value (e.g.: R$6000, R$500, ...) breaks the sorting. With this minor change, brazilian currency can be supported and I didn't see any other impact while using the modified version.

### DIFF
--- a/docs/option/columns.type.xml
+++ b/docs/option/columns.type.xml
@@ -21,11 +21,12 @@
       * Sorting - sorted numerically
       * Filtering - no effect
     * `num-fmt` - Numeric sorting of formatted numbers. Numbers which are formatted with thousands separators, currency symbols or a percentage indicator will be sorted numerically automatically by DataTables.
-      * Supported built-in currency symbols are `$`, `£`, `€` and `¥`.
+      * Supported built-in currency symbols are `$`, `£`, `€`, `¥` and `R$`.
       * Supported built-in thousands separators are `'` and `,`. 
       * Examples:
         * $100,000 - sorted as `100000`
         * £10'000 - sorted as `10000`
+        * R$1000 - sorted as `1000`
         * 5'000 - sorted as 5000
         * 40% - sorted as 40
       * Sorting - sorted numerically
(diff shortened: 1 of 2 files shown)


DataTables:
Thanks for the PR.

The reason I didn't include this originally is that it is just the letter `R` being added to the regex. The result is that if a column contains the letter `R` alone it could be detected as currency incorrectly.

I'm not sure about pulling this in for that reason - it is an edge case an unlikely, but it could break things unexpectedly. I'll need to have a think about including it in the core, rather than expecting a plug-in to be used, as currently is the case.

On the plus side there are a number of other currencies that also use `R$` for their currency which would automatically gain support from this change.

rodrigoberriel:
@DataTables I changed the regex a little bit and now it detects as currency only if "R" is followed by one "$".

Tests:
```
dataTable({
    language: {
        "decimal": ",",
        "thousands": ".",
        url: 'datatable/pt_BR.json'
    }
})
```
- ```$``` symbol only (detected as currency):
image

- ```$``` and ```R$``` symbol (detected as currency): 
_just for test, I don't think people would mix different currencies in one table_
image

- ```$```, ```R$``` and ```R``` symbol (detected as text, ```R``` breaks currency detection as expected):
image

If you think it should be implemented in another way, feel free to close the PR.
**Thanks for the awesome jQuery plug-in you've made!**

DataTables:
I've just updated DataTables to include support for the top 20 internationally traded currencies (which includes the Real :-) ).

To address my concern above about data points being incorrectly classified I've changed DataTables formatted number type checker slightly so if a string without numbers is passed in that passes the regex, it will no longer be detected as a formatted currency (so in this case `R`, `R$`, `kr`, etc would all be classified as strings, unless they also have a number).

An additional 32 bytes to the library size, but the currency support is fairly complete now I think! Any other currencies I think I will insist on being in a plug-in.

Thanks for bringing this up.

PR #40, opened 2015-04-24
Add a pre-filter callback as a configuration option

tcrossland:
I've added a callback configuration option to allow the filter term to be preprocessed. The obvious use case is to remove accents without having to bind to the filter input's keyup event (see http://www.datatables.net/plug-ins/filtering/type-based/accent-neutralise).

### DIFF
--- a/js/core/core.constructor.js
+++ b/js/core/core.constructor.js
@@ -141,6 +141,7 @@ _fnMap( oSettings, oInit, [
 	"iTabIndex",
 	"fnStateLoadCallback",
 	"fnStateSaveCallback",
+	"fnPreFilterCallback",
 	"renderer",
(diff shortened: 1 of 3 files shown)


tcrossland:
Contribution is offered under and will be made available under the project's existing license (MIT).

DataTables:
Hi,

Thanks very much for the PR. I think at the moment, what I would prefer is that the DataTables API will take the string to be search and use it directly, rather than modifying it though. If modification is required, then it should be done before calling the `search()` method.

I see the convenience of doing it the way you have, but at the moment it isn't something I want to include in the core.

One point about the PR - everything you've done is spot on, thank you! The one thing I would say is that the column filters would need something similar, if not identical.

PR #41, opened 2015-04-24
Remove label elements to improve compatibility with bootstrap input-groups

tcrossland:
Bootstrap's [input groups](http://getbootstrap.com/components/#input-groups) can be nice for filtering and selecting, but Datatables' hard-coded labels don't allow for this.

 * [Currently](https://jsfiddle.net/tomxland/2st8bxma/1/embedded/result/)
 * [After this patch](https://jsfiddle.net/tomxland/fmwt7znx/1/embedded/result/) - visual differences only as I can't pull the patched datatables js from the CDN

### DIFF
--- a/js/core/core.filter.js
+++ b/js/core/core.filter.js
@@ -23,7 +23,7 @@ function _fnFeatureHtmlFilter ( settings )
 			'id': ! features.f ? tableId+'_filter' : null,
 			'class': classes.sFilter
 		} )
-		.append( $('<label/>' ).append( str ) );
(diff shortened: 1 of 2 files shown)


tcrossland:
Contribution is offered under and will be made available under the project's existing license (MIT).

DataTables:
Thanks for the PR.

The `label` element is used to help improve accessibility for assistive devices. Removing them is not currently an option I'm afraid. You could use a callback such as `initComplete` if you need to remove them.

PR #42, opened 2015-05-14
Add preRowPopulate callback option

cdaringe:
# problem statement
I source data for a table from a huge array.  This array represents a `Collection` of `Model`s.  When each [`Model`](https://github.com/AmpersandJS/ampersand-model) is instantiated, it become quite a data rich object.  If I instantiate the each Model in the array immediately, the browser will choke when the array is large.  Instead, for performance reasons, if I can mutate the DT row data immediately before column population, I can do mutations (i.e. create my data-rich Models) **just-in-time** and only worry about processing those items that DataTables is rendering!

# solution
mimic `createdRow`'s implentation, as it's purpose is nearly the same, but should be fired *before* the row is populated, not after.

# review
Any docs/etc that may need updating, I am happy to do.  Just give me a boost with some pointers! :+1:

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -16,7 +16,7 @@ This separation allows developers who simply want to use DataTables as is, to do
 
 ## Building
 
-DataTables can be built using the `make.sh` script in the `/scripts` directory of this repo. Simply check out the repo and run `make.sh --help` to get a full list of the options available for the build process. `make.sh build` will be the most common (with `make.sh build debug` available for quick testing - it skips the minification steps for speed).
+DataTables can be built using the `make.sh` script in the `/build` directory of this repo. Simply check out the repo and run `make.sh --help` to get a full list of the options available for the build process. `make.sh build` will be the most common (with `make.sh build debug` available for quick testing - it skips the minification steps for speed).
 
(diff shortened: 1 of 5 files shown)


DataTables:
Hi,

Thanks for the PR. Are you able to give me an example of how this would work please? I'm struggling to see how it would if the defer render option was enabled, since all the models would be required up front anyway to access the data.

cdaringe:
Hi Allan:

I need to revisit my strategy.  I started typing this out, but saw its flaw.

----
I can make a fiddle, but I think perhaps some more explanation would help.  The issue is I need to preprocess the data before rendering it, and I don't want to blindly pre-process everything.

```js
var arr = [{x: 1}, {x: 2}, ... {x: 1e7}];
for (var i = 0; i < arr.length; ++i) {
   // maybe add a y attr
   arr[i].y = 'y';
  // and then some **computationally heavy lifting**, perhaps each cycle takes 1ms
  arr[i] = computationallyHeavyMutation(arr[i]);
}

// i've now just waited at least 1e7 ms to pre-process the data
```

instead, I want to pre-process only the rows that DT intends on rendering:

PR #43, opened 2015-05-18
jq 2.1.4 sppt

cdaringe:
The problem statement for this issue is described here:
https://github.com/DataTables/DataTables/issues/557
This was the only other spot found using `$.map` in the vulnerable way


DataTables:
Very good catch - thank you! Are you able for this to be included in DataTables under the MIT license?

cdaringe:
absolutely.  does that require any further follow up?

DataTables:
Nope - that's perfect. Thanks!

PR #48, opened 2015-07-10
fix for incorrect calculation in _fnBrowserDetect

ieaglle:
when body's position is changed (left: 375px) _fnBrowserDetect
incorrecly calculates offsets. thus header is a bit misaligned. commit
changes test object's position to fixed. in this case it's position will
be relative to the viewport not to the body.
this fixes [#594](https://github.com/DataTables/DataTables/issues/594)

I agree that the change will be released under the MIT license.


DataTables:
Nice one - thanks! I'll have this in the 1.10.8 release.

PR #49, opened 2015-07-14
Fix: server-side columns().search() call passes empty string to _fnFi…

zeigotaro:
…lterComplete

Server-side issue, difficult to set up test page using publicly
available sAjaxSource that handles DataTables-specific parameters

### DIFF
--- a/js/api/api.search.js
+++ b/js/api/api.search.js
@@ -50,7 +50,12 @@ _api_registerPlural(
 				"bCaseInsensitive": caseInsen === null ? true : caseInsen
 			} );
 
-			_fnFilterComplete( settings, settings.oPreviousSearch, 1 );
+            _fnFilterComplete( settings, $.extend( {}, settings.oPreviousSearch, {
+                "sSearch": input+"",
+                "bRegex":  regex === null ? false : regex,
+                "bSmart":  smart === null ? true  : smart,
+                "bCaseInsensitive": caseInsen === null ? true : caseInsen
+            } ), 1 );
 		} );
 	}
 );


DataTables:
Could you clarify what issue this change fixes please?

Using [this example](http://datatables.net/examples/server_side/simple.html) if I open the console and enter:

    $('#example').DataTable().column(0).search('Airi').draw()

It correctly reduces to a single row. Then:

    $('#example').DataTable().column(0).search('').draw()

correctly removes the filter.

zeigotaro:
I'm running into an issue where, with aoColumnDefs set, settings.oPreviousSearch does not have the sSearch field set correctly. Changing that call to _fnFilterComplete to match the call used for _api_register( 'search()'...) (the function above) fixes the issue I'm seeing.

DataTables:
Are you able to modify [this server-side processing example](http://live.datatables.net/melizeya/1/edit) to demonstrate the issue please. I don't see why defining `columnDefs` would make any difference to the column search API.

zeigotaro:
Thank you for setting that example up. I've modified it, but the behavior I'm seeing on the JSBin is different than what I'm encountering  (I'm using a JsonResult created with the DataTablesParser in the DataTablesDotNet library). I can workaround on my end by uncommenting that "oSettings.oPreviousSearch.sSearch" assignment. If you do not feel this is a productive fix, feel free to close. Thanks for your help. :-)

DataTables:
Have you got a link to the DataTablesDotNet library? I wonder if it is a bug there.

I'd like to take the fix, but before I do I need to understand what the bug is :-).

zeigotaro:
https://github.com/timhobbs/DataTablesDotNet is where that library is located. Thanks!

DataTables:
Great thanks. I'll try to take a look before the 1.10.8 release drops.

PR #50, opened 2015-07-21
Resolve issue with focusing when there are multiple graphs updating on the same page

jfktrey:
Trivial fix. Issue arises when there are multiple datatables on a single page and both of them are updating via `.draw(false)` to hold their position. Having a page link on the first graph focused with the tab key when the draw call is made causes the selected page to jump to the second graph. In the case of bootstrap, clicking a page link will cause the element to remain focused, so this case arises often.

You can see this in action [here](http://codepen.io/anon/pen/zGJzbN). Click in the search box, then tab until a page link (1 or 2) is selected. Focus will jump to the corresponding link on the second graph.

The primary issue is when the second graph is below the bottom of the screen, causing jarring scrolling.

### DIFF
--- a/js/ext/ext.paging.js
+++ b/js/ext/ext.paging.js
@@ -158,7 +158,7 @@ $.extend( true, DataTable.ext.renderer, {
 				// elements, focus is lost on the select button which is bad for
 				// accessibility. So we want to restore focus once the draw has
 				// completed
-				activeEl = $(document.activeElement).data('dt-idx');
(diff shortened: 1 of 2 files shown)


DataTables:
Very good point and nice fix. Thanks for the PR!

Are you happy for this to be released under the MIT license?

jfktrey:
I am happy with this code being released under the MIT license.

Thanks for the fast response!

DataTables:
Awesome - thank you :-). The fix will be in 1.10.8 which should be available within the next two weeks.

PR #53, opened 2015-08-14
Prevent style undefined error

NoMan2000:
Throws a style undefined error when it cannot find the header cell, this prevents the error.  This occurs when using ColVis to hide/show certain rows after the page has loaded.

This fork is made under the MIT license and you are free to do as you wish.

### DIFF
--- a/js/core/core.sizing.js
+++ b/js/core/core.sizing.js
@@ -86,7 +86,9 @@ function _fnCalculateColumnWidths ( oSettings )
 
 		for ( i=0 ; i<visibleColumns.length ; i++ ) {
 			column = columns[ visibleColumns[i] ];
-
+			if ( !headerCells[i] ) {
+		        	continue;
+	                }
 			headerCells[i].style.width = column.sWidthOrig !== null && column.sWidthOrig !== '' ?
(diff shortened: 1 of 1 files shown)


DataTables:
Can you show me some example HTML that would trigger this issue please?

NoMan2000:
http://g.recordit.co/IhqBcAu8tN.gif

Here's a live example, showing the code would be difficult to explain, but it's using ColVis plugin and changing from hidden/visible.  Hiding works okay but then when showing the rows again error occurs.  

Plus as a general defensive coding practice I can't see anything wrong with checking that a variable exists before attempting to use it.

DataTables:
I agree in principle, but my point is that there is something else going wrong here and I'm concerned that this will just mask it. That error shouldn't occur, so I would like to understand why it is and if a larger change is needed.

Also what version of DataTables are you using?

NoMan2000:
Mea culpa, I thought I had pulled the latest version but the old one was still cached in the browser when I did the test.

With the latest version, the problem disappeared.  Seems to have fixed a few other issues as well.  Ignore/Commit this patch as you see fit.

DataTables:
Good to hear that the latest version fixed this! :-)

PR #54, opened 2015-08-17
Update rows().ids().xml

gauravmak:
Spell mistake


DataTables:
Thanks - there are actually a few like this. I'll need to grep through the docs and fix them up and put a redirect in on the server.

I should just check, are you happy for this change and your others to be released under the MIT license?

gauravmak:
Of course.

DataTables:
Lovely - thanks :-)

PR #55, opened 2015-08-29
TypeError: val is null in for ( var j=0, jLen=val.length ; j<jLen ; j++ )

neoglez:
The function `-api _fnSetObjectDataFn( mSource )` returns a function that can be used to set data from a source object, nevertheless it doesn't handle correctly the case when the value is null `(val === null)`, which is a legitimate value. This can cause problems for function relaying on this function. One example is when Editor use this function to set the values `-api Editor.prototype._submit`, if a user didn't select anything in a multiple value select, the value for this field is null. A very simple way (but probably not the most elegant) to solve this one is by ignoring a null value passed an early returning.

### DIFF
--- a/js/core/core.data.js
+++ b/js/core/core.data.js
@@ -345,6 +345,11 @@ function _fnSetObjectDataFn( mSource )
 	{
 		/* Like the get, we need to get data from a nested object */
 		var setData = function (data, val, src) {
+			// early return on null values
+			if (val === null)
+			{
+				return;
+			}
 			var a = _fnSplitObjNotation( src ), b;
 			var aLast = a[a.length-1];
 			var arrayNotation, funcNotation, o, innerSrc;
--- a/js/core/core.sizing.js
+++ b/js/core/core.sizing.js
@@ -317,6 +317,10 @@ function _fnStringToCss( s )
 			'0px' :
 			s+'px';
 	}
+	
+	if (typeof s.match == 'undefined') {
+		return null;
+	}
 
 	// Check it has a unit character already
 	return s.match(/\d$/) ?


DataTables:
Hi - thanks for the PR. It looks like it doesn't actually set null values in the data source object now. If the value is null, then that should be set in whatever data store is being used, and not just ignored as it appears to be with the PR (as you say it is a legitimate value).

Is the issue you are seeing that you are getting a warning from DataTables about an unknown parameter? If so, you need to use the `columns.defaultContent` option to tell DataTables how to "display" a null value. Normally this is an empty string, but you could also have "Not set", etc.

neoglez:
Unfortunally setting columns.defaultContent doesn't help.

1.Press New.

2.Watch the console ;)

http://live.datatables.net/lubasaga/1/edit

The question is where do you want to allocate the responsability of handling null values: in function ` _fnSetObjectDataFn( mSource )` or in the components calling this function (like Editor). I say ` _fnSetObjectDataFn( mSource )` should really be able to set null value on nested sources.

DataTables:
Thanks for the test case. The issue isn't so much that DataTables can't handle null values at all, but rather specifically it can't where it has been told to expect an array using the array notation in the rendering function.

I'm surprised jQuery returns null rather than an empty array, when it returns an array when items are selected - that seems counter-intuitive to me - I've opened [an issue](https://github.com/jquery/jquery/issues/2562) to see if someone would be wiling to explain why it is done this way.

However, it is done that way, so I've modified DataTables to allow for this.

Thanks for the PR, however, I haven't pulled it in as it actively disables the ability to set a null value at all (regardless of the array notation used). The value (even `null`) should be written to the data source object, which doesn't happen in your code.

neoglez:
https://github.com/DataTables/DataTablesSrc/commit/4aee9730f2ab77f7b8caf07bad95b002b0b3710b looks like a really robust fix. :clap:

PR #56, opened 2015-09-10
Added a 'custom' method.

Masquerade-Circus:
Added a 'custom' method to support more complex queries to the database. 

You could use Joins or group functions like sum, min, max, count, etc.

### DIFF
--- a/examples/server_side/scripts/ssp.class.php
+++ b/examples/server_side/scripts/ssp.class.php
@@ -16,6 +16,13 @@
  */
 
 
+// REMOVE THIS BLOCK - used for DataTables test environment only!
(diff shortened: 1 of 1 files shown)


DataTables:
Thanks for the PR. Let me have a little think about this one -  really want this to be a demo script only and super simple at that, while this adds a fair amount of complexity.

How does it handle column searching from the columns specified in the select string - presumably it can't?

Masquerade-Circus:
I worked on these features and now it can do it. Please see the example in the comments blocks of the functions.

DataTables:
Thanks - I'll try to review it as soon as possible (although I've got a significant backlog atm, so it might be a little while before I can).

PR #57, opened 2015-09-11
The change discussed in #653

strang91:
Moved from https://github.com/DataTables/DataTables/pull/664

### DIFF
--- a/js/api/api.cells.js
+++ b/js/api/api.cells.js
@@ -54,24 +54,10 @@ var __cell_selector = function ( settings, selector, opts )
 		return allCells
 			.filter( s )
 			.map( function (i, el) {
-				if ( el.parentNode ) {
-					row = el.parentNode._DT_RowIndex;
-				}
-				else {
-					// If no parent node, then the cell is hidden and we'll need
-					// to traverse the array to find it
-					for ( i=0, ien=data.length ; i<ien ; i++ ) {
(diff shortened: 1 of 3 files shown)


DataTables:
Thanks for this - are you happy for this to be included under the MIT license?

I don't have a coding standard document published, but I do always require braces be used. I can change the `continue` (I would actually prefer it to be negated so there is no continue actually) or if you would, that would be great.

strang91:
MIT License is good for me.

No problem, Just tell me what you wish to be changed & I'll change it...

DataTables:
The logic for the `continue` could be inverted, so rather than having a continue, just check for the cells before using them. I'll try running it, probably on Monday and run unit tests on it and hopefully commit it in then.

strang91:
Replaced the continue with an if condition.

DataTables:
I'm afraid I've not had a chance to look into this yet - but I've not forgotten. I will do asap.

strang91:
Hi. Have you had a chance to look into this? Is there a chance that this patch will make it into the next release?

Thanks again.

DataTables:
I plan to review it this coming week :-)

DataTables:
Merged in now - thanks for your input with this - hugely appreciated! It will be in 1.10.10 which will be released later this week or next.

strang91:
Thanks

PR #58, opened 2015-09-25
Preventing re-sort and re-filter when updating data

montrothstein:
We have as use case where users are working through a list of items in a table.  They may have applied a filter and/or sort to the table.  As they work through the list changes they make can be to values used for sorting or filtering.

We want the changes to appear in the table but we don't want them to loose their place in the list.

To Address this we added a "hold-order" option to draw() which in turn enables a ignoreSortAndFilter flag on _fnReDraw().  This disables sorting and filtering therefore achieving the desired effect.

I don't know how useful this would be to others but I thought I'd contribute it.

### DIFF
--- a/js/api/api.draw.js
+++ b/js/api/api.draw.js
@@ -8,6 +8,11 @@ _api_register( 'draw()', function ( paging ) {
 		if ( paging === 'page' ) {
 			_fnDraw( settings );
 		}
+		// To update the data without re-sorting, re-filtering, or changing the page, the hold-order option was added.
+		// This sets the ignoreSortAndFilter falg on _fnReDraw.
+		else if (paging === 'hold-order') {
+					_fnReDraw(settings, true, true);
+		}
 		else {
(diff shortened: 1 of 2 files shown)


DataTables:
Does the existing `page` option not do what you need?

montrothstein:
I failed to mention in my writeup that we need to handle inserts as well as edits.  The 'page' option shows edits but not inserts.

Obviously inserts (or deletes) can cause a shift in position.  If there are multiple at once then the shift could be significant.  However our inserts are all singular and a slight shift (even if the currently selected row moves to the next page) is less disorienting than a full re-sort/filter.

DataTables:
Thanks for the extra info. Yes, it sounds like there might be a couple of issues that would need to be catered for in the general case - I'm afraid I don't have the time too look into it in detail just at the moment, but I will leave this one open until I do (which hopefully won't be too far away!).

PR #60, opened 2015-12-15
Added check to ensure the sorting column exists

0xbrock:
Attempting to not have a default sort causes an issue where it assumes there will always be a sort column. My pull request fixes this issue.
Example site: http://0xbrock.github.io/DataTablesSrc/
From FireBug: 
TypeError: aoColumns[srcCol] is undefined
aDataSort = aoColumns[ srcCol ].aDataSort;

From Chrome:
Uncaught TypeError: Cannot read property 'aDataSort' of undefined

### DIFF
--- a/js/core/core.sort.js
+++ b/js/core/core.sort.js
@@ -41,7 +41,7 @@ function _fnSortFlatten ( settings )
 	for ( i=0 ; i<nestedSort.length ; i++ )
 	{
 		srcCol = nestedSort[i][0];
-		aDataSort = aoColumns[ srcCol ].aDataSort;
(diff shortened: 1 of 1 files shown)


DataTables:
> "order": [[]],

This isn't a valid value for the `order` option. If you don't want a default order just use `order: []`. If you specify inner arrays then they _must_ contain the two elements (column and direction).

See also the [documentation for `order`](http://datatables.net/reference/option/order) which contains an example that has do default order.

0xbrock:
I thought I had tried that to no avail, but upon retrying it works.  Thanks!

PR #61, opened 2015-12-25
Doc updates

adilapapaya:
First off, thank you for such an amazing library! 

I made a few small changes to issues I came across while trying to build from source:

* `Contributing.md`: Added a link to the [DataTables/DataTablesSrc](https://github.com/DataTables/DataTablesSrc) and  [DataTables/DataTables](https://github.com/DataTables/DataTables) repos in Contributing.md file. (It makes it a little clearer that DataTablesSrc is a repo and not a directory.) Note: may need to update the `Contributing.md` file in the [DataTables/DataTables](https://github.com/DataTables/DataTables) repo as well.

* `Readme.md`: Updated the instructions for building the project
    * Added links to some of the external programs so it's easier to get started installing.
    * Added note regarding where to install these programs.
    * Added important note about how things will get overwritten if you don't commit first. (*I'm not sure how much exactly is overwritten...just that every time i makde a change to the `Readme` and didn't commit, it would overwrite it with what was in the repo.*)

* `make.sh`: Change from using `git clone git@...` to `git clone https://...`. I think github changed something as the `git@` version doesn't work anymore.

--------

*Contributed under the MIT license*

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -7,7 +7,7 @@ Please direct support requests to the [DataTables forums](https://datatables.net
 
 If you are thinking of contributing code to DataTables, first of all, thank you! All fixes, patches and enhancements to DataTables are very warmly welcomed. In order to keep thing manageable, there are a number of guidelines that should be followed in order to ensure that your modification is included in DataTables as quickly as possible:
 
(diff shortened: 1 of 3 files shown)


DataTables:
Thanks for this - very much appreciated :-)

PR #63, opened 2016-01-07
enable js createlement in render function

icepower03:
this change allows to make this ( the click event in particular)

// extract from the datatable column definition in full js
 {
              title: "accès bureau de prescription",
              render: function (data, type, row, meta)
              {
                  return $('<div/>', {
                      text: "Presc. V5",
                      class: 'ouvrirPrescV5'  
                  })
                .click(function () {  lancerBureau(row); })[0];
              }
          },

i think that the click event should be chained and not triggered with a new selector on tbody which to recreate the data() object.

so i can make it work like that but i think that it's not logical /or/ the most efficient way
  .on('click', 'tbody .ouvrirPrescV5', function () {

```
     var dt = monTableau.row($(this).parents('tr')).data();
     lancerBureau(dt);
  });
```

//the idea is to full manage the contenent of the cell including events
// if you have another way to do such, this change is useless but as i see that the innerHTML is used to insert content in td i think that only dom strings are accepted and not dom objects.

### DIFF
--- a/js/core/core.draw.js
+++ b/js/core/core.draw.js
@@ -48,7 +48,14 @@ function _fnCreateTr ( oSettings, iRow, nTrIn, anTds )
 			// Need to create the HTML if new, or if a rendering function is defined
 			if ( !nTrIn || oCol.mRender || oCol.mData !== i )
 			{
-				nTd.innerHTML = _fnGetCellData( oSettings, iRow, i, 'display' );
+			   	var cellData = _fnGetCellData(oSettings, iRow, i, 'display');
+			   	// if the render function of columns creates directly a dom object
+                    		if (typeof cellData === 'object')
+                    		{
+                        		nTd.appendChild(cellData);
+		                }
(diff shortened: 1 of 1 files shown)


DataTables:
Thanks for the PR. The issue with this is that it might break the extensions such as Buttons etc, which don't expect a jQuery object. It could also cause issues for type detection and filtering, which again, don't expect a jQuery object.

I'm afraid I don't think I can pull this in at this time.

Having said that, being able to return a jQuery object (or DOM node) is something I plan to add in the next major version of DataTables - but its going to take some careful thought.

icepower03:
ok thanks for your reply, i let you think about it.

milux:
This is basically a less powerful approach to the technique I suggested over one your before you came up with your suggestion, see advanced rendering #19

milux:
Plus - as far as I can tell - my approach doesn't suffer from the problem mentioned above.

PR #65, opened 2016-01-22
saveState: correction to documentation, and new example

frank-carnovale:
I can see why there are issues about unpredictable behaviour of saveState.  The documentation says that state is saved only against the table's DOM id but the code silently extends this id with the current page pathname.

So by default in an app that uses a REST-style path scheme the state is saved for e.g. "/view/group/10" but is not available for "/view/group/11".

This PR clarifies the documentation and adds an example which shows how to save and load localStorage against the DOM id-only; which may be more appropriate for some REST-style apps.

This contribution and any further contributions by myself (frank.carnovale@gmail.com) is offered under and will be made available under the DataTable project's existing license (MIT).

### DIFF
--- a/docs/option/stateSave.xml
+++ b/docs/option/stateSave.xml
@@ -8,11 +8,11 @@
 	<default value="false" />
 
 	<description>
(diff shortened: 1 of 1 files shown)


DataTables:
Awesome - thanks for this!

PR #67, opened 2016-02-05
Check if index of aoColumns resolves to an object

JimHough:
I was getting an exception with scrollX enabled here because the index doesnt resolve to an object.
I think it could be because I am adding another row to the header to contain per-column search boxes.

### DIFF
--- a/js/core/core.scrolling.js
+++ b/js/core/core.scrolling.js
@@ -229,7 +229,8 @@ function _fnScrollDraw ( settings )
 
 	$.each( _fnGetUniqueThs( settings, headerCopy ), function ( i, el ) {
 		idx = _fnVisibleToColumnIndex( settings, i );
-		el.style.width = settings.aoColumns[idx].sWidth;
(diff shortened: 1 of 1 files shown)


DataTables:
Could you link to a page showing the issue please? There shouldn't be any issue with having to header rows in a scrolling table - as shown in this little example: http://live.datatables.net/pibepana/1/edit .

Also, could you indicate if you are happy for the to be merged in under the MIT license if I do pull it in?

JimHough:
I'm still trying to pin down what is the root cause of the issue.
I think it is because I'm dynamically hiding and showing columns but cant reproduce the error in isolation of my app.
I will try to make a minimal reproduction of the issue and have no issue with you merging the change under the MIT license.

DataTables:
Going to close this form the moment as I haven't been able to reproduce the issue and haven't had any other reports of it. If this is still a problem, a test cases showing it would be great.

PR #75, opened 2016-04-01
Fix: Bootstrap 4 flexbox grid

stroebjo:
The flexbox enabled grid in bootstap 4 requires the lowest `.col-` to be set. https://github.com/twbs/bootstrap/issues/17603#issuecomment-180862254

### DIFF
--- a/js/integration/dataTables.bootstrap4.js
+++ b/js/integration/dataTables.bootstrap4.js
@@ -46,9 +46,9 @@ var DataTable = $.fn.dataTable;
 /* Set the defaults for DataTables initialisation */
 $.extend( true, DataTable.defaults, {
 	dom:
-		"<'row'<'col-md-6'l><'col-md-6'f>>" +
-		"<'row'<'col-md-12'tr>>" +
-		"<'row'<'col-md-5'i><'col-md-7'p>>",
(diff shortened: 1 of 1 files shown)


DataTables:
I'm afraid I don't quite understand. If the viewport is less than `md` then I want the columns to be broken up into rows. Has that changed in v4? The [docs don't seem to reflect such a change](http://v4-alpha.getbootstrap.com/layout/grid/).

stroebjo:
In the classic float grid it works in v4, but when you enable flexbox in Bootstrap the rows wont be 100% width. See the attached screens. 
If you use the flexbox grid the grid cells dont fallback to 100% width, on lower breakpoints, you need to explictly define die lower breakpoints (`col-xs-12`). 

Classic float grid in v4, centered cells on xs.
twbs4-float-col-md

Flexbox grid, without `col-xs-12`, rows stick to left. 
twbs4-flex-col-md

Flexbox grid, added `col-xs-12`, works as expected.
twbs4-flex-col-xs

DataTables:
Ah I see - thanks for the clarification. I wonder if we should wait for the BS issue you linked to to be resolved before making changes here.

DataTables:
Sorry its taken me so long to get around to this! I agreed, based on mdo's comment's in that bug that your fix is correct. Thanks!

PR #79, opened 2016-06-27
Fix PHP warnings

shmax:
PHP has moved away from the practice of mirroring constructor names with the class name. This quells a lot of angry warnings for people trying to build with a more recent version of PHP, but doesn't break 5.4, which is the current build dependency.


DataTables:
Fantastic - thanks. Are you happy for these changes to be released under the MIT license?

shmax:
Of course. Merge away.

PR #80, opened 2016-07-06
Fix date sorting of empty fields

aij:
Previously, date sorting of values that
failed to parse as a date were sorted as
if they were 1970-01-01 (Unix and JS epoch).

Now, they sort before other dates.

Example: https://jsfiddle.net/qae6zjty/1/

I acknowledge that my contribution is offered under and will be made available under the MIT license.


DataTables:
Agreed - thanks very much for this! I'll merge it tomorrow when I get a chance to write this into a unit test.

DataTables:
Tests for this specific case added here: 69b7a455668649d21b2be77e1304859aa3a8a55d . We are in the process of adding more tests... :-)

Thanks for this PR!

Issue #81, opened 2016-07-15
Datatables _fnSetObjectDataFn function does not store array types correctly


jimjaeger:
Fix for _fnSetObjectDataFn required to handle storing of array types correctly.
Already open a forum entry. Also I already fixed it on a local branch. If you give me rights for branch upload I can provide the fix to you.

+ arrayNotation = aLast.match(__reArray);
+var join = arrayNotation[0].substring(1, arrayNotation[0].length-1);
+var r = val.length > 0 ? val.split(join) : [];
+ data[ aLast.replace(__reArray, '') ] = r;
- data[ aLast.replace(__reArray, '') ] = val;

DataTables:
Thanks. Could you give me a link to the forum thread please? Does it give details on example data showing the issue as it was?

Pull requests can be sent to the master branch here and I'll consider them for inclusion.

Unit tests are something that DataTables is currently lacking for these methods, although they are inbound.

DataTables:
[I found it here](https://www.datatables.net/forums/discussion/36289/datatables-fnsetobjectdatafn-function-does-not-store-array-types-not-correct?new=1).

If you could please link to a test case showing the issue that would be great.

jimjaeger:
Hi,

here is an example of data for testcase:
{"total_rows":104,"offset":0,"rows":[ {"id" : "myid",	"doc" : {"tags" : ["tag1"],"_id" : "myid",	"_rev" : "3-188272"	} }]


sample table config:

var table = $('#dataTables-projectmetadata').DataTable({
		responsive : true,
		paging : true,
		fixedHeader : true,
		stateSave : true,
		sAjaxDataProp: 'rows',
		rowId: 'id',
		"ajax": "/api/metadata",
		"columns": [
		            { "title": "ID", "data": "id" },
		            { "title": "Tags", "data": "doc.tags[, ]", "defaultContent": "" }
		        ]
	    });


After I modify the entry  the new value is now {"id" : "myid",	"doc" : {"tags" : "tag12","_id" : "myid",	"_rev" : "3-188272"	} }

from array to string ....

I guess I have to create a branch and send a pull request after my changes, but I have not access rights to create a branch.

DataTables:
Thanks! Let me look into it, but in the short term I would recommend using:

```
{ "title": "Tags", "data": "doc", render: 'tags.[, ]', "defaultContent": "" }
```

That takes the array setter out of the equation.

jimjaeger:
[...]

PR #83, opened 2016-07-24
Update core.data.js

jimjaeger:
Fix array value handling in _fnSetObjectDataFn function to correct store array notation value in array instead of string.


jimjaeger:
Fix for issue #81

PR #84, opened 2016-09-30
Use faster [].filter instead of in-place [].splice in filter functions

kkopachev:
Use case: Given table with 50 columns and 20k rows. Splice executes ~5s, using filter gives less then a second response.

I am happy for these changes to be released under the MIT license.

### DIFF
--- a/js/core/core.filter.js
+++ b/js/core/core.filter.js
@@ -189,16 +189,28 @@ function _fnFilterColumn ( settings, searchStr, colIdx, regex, smart, caseInsens
 		return;
 	}
 
-	var data;
-	var display = settings.aiDisplay;
 	var rpSearch = _fnFilterCreateSearch( searchStr, regex, smart, caseInsensitive );
 
-	for ( var i=display.length-1 ; i>=0 ; i-- ) {
-		data = settings.aoData[ display[i] ]._aFilterData[ colIdx ];
+	if ( Array.prototype.filter ) {
+		settings.aiDisplay = settings.aiDisplay.filter( function ( v ) {
+			return rpSearch.test( settings.aoData[ v ]._aFilterData[ colIdx ] );
+		} );
+	}
+	else {
+		// Compatibility for browsers without EMCA-252-5 (JS 1.6)
+		var display = settings.aiDisplay;
+		var newDisplay = [];
+		var data;
 
-		if ( ! rpSearch.test( data ) ) {
-			display.splice( i, 1 );
+		for ( var i = 0, ien = display.length; i < ien; i++ ) {
+			data = settings.aoData[ display[i] ]._aFilterData[ colIdx ];
+
+			if ( rpSearch.test( data ) ) {
+				newDisplay.push( display[i] );
+			}
 		}
+
+		settings.aiDisplay = newDisplay;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


DataTables:
Hi,

Thanks for this! Unfortunately `Array.prototype.filter` isn't [supported in IE8](http://kangax.github.io/compat-table/es5/#test-Array.prototype.filter). In fact, DataTables core currently supports to IE6, although with the next non-patch release that will change to IE8+.

Given the performance benefit it will be worth looking into having conditional code here, so I'm not going to close this PR, but rather keep it open until I have time to look into that properly.

Sorry I can't pull it in as is.

kkopachev:
@DataTables Thanks for your response. I can add conditional code or add polyfill, whatever your preference is.

DataTables:
Personally I would prefer a polyfill, but I don't think it would be a good idea for the library to add that since it could potentially impact others. Conditional code would be the way to go here.

kkopachev:
@DataTables code updated. 
I tried to compare splice with for loop with array push. Array push works ~35 times faster on my machine, so decided to do that instead of splice.

I tried to extract duplicated code, but it tends to be either too complex for the task, or has too generic name for specific usage. 

Please let me know what you're thinking.

DataTables:
Can you show me your code for array push? I'm not clear how that would work when the goal is removing an item from the array. Are you basically creating a new array in the loop and then assigning that as the display array?

My one concern about that is that `aiDisplay` is then no longer the original array. I'm not sure that is an issue (I doubt it, but I'm not _certain_).

kkopachev:
@DataTables code is in that PR. I added conditional to use `Array.prototype.filter` if available, otherwise create new array and push valid items into it.

As of your concern, looks like it is never used as a reference. In fact, `aiDisplay` in lot of places are created just as a shallow copy of `aiDisplayMaster`  with calls like this:
```js
/* Copy the data index array */
oSettings.aiDisplay = oSettings.aiDisplayMaster.slice();
```
I could not find any place where reference is explicitly used.
There is one place which might be affected by this change - in header/footer callbacks. Also other callbacks where settings are passed as parameter. That appears very strange for me for someone to change display array from inside callback and looks like at the moment header/footer callbacks fired, changing `aiDisplay` array would not have any visual results.

However, there is commit https://github.com/DataTables/DataTablesSrc/commit/f07529f7ddf780f9668356727be62df70b3da153
which does filtering creating new array, but then backmerging it into original array. I can add that piece of code too, if there is still concerns.

DataTables:
Sorry - yes I see what you mean now. So are you saying that `push` works faster than either `filter` or `splice`? That would be good news as it would support everything needed.

I also think you are right about the reference not needing to be maintained. I need to check over my code to convince myself that is actually true, but I'm reasonably certain it is!

kkopachev:
Hm, I was trying to say that `push` is faster then `splice`, but `filter` is faster then `push`. But did some more testing and it looks like `push` might be faster in some runs. :confused: 
In Chrome `push` seems to be a little bit faster then `filter`, but very close. `filter` calls fluctuate more.
In FF - `push` is 75x slower then `filter`.

Here is my testing code:

```js
var N = 100000;
var q = Array.apply(null, {length: N}).map(Number.call, Number);
var q1 = q.slice();
var q2 = q.slice();
var q3 = q.slice();


console.time('a');
var a = [];
for ( var i = 0, ien = q1.length; i < ien; i++ ) {
    if (q1[i] % 2 == 0) {
        a.push( q1[i] );
    }
}
console.timeEnd('a');
delete a;

console.time('b');
for (var i = q2.length - 1; i >= 0; i--) {
    if (q2[i] % 2 == 0) {
        q2.splice(i, 1);
    }
}
console.timeEnd('b');


console.time('c');
var c = q3.filter(function(v){ return v % 2 == 0 });
console.timeEnd('c');
```

So I'd probably keep both `filter` and `push` as fallback for older browsers.

DataTables:
I'm inclined to use just the push method since that is most compatible and it reduces code bloat (one implementation rather than two - DataTables is already larger that I would like!).

I'll play around with it a bit! Thanks for your insights into this :-)

PR #86, opened 2016-10-26
Allow templating in sortAscending and sortDescending ARIA strings

ferranmesas:
ARIA labels for sortable columns are built using raw string concatenation. With this pull request, the same string templating that works in other components (like in pagination: `Showing page _PAGE_ of _PAGES_`) can be used for both of these strings.

I offer this contribution under the same license the project uses, the MIT license.

### DIFF
--- a/docs/option/language.aria.sortAscending.xml
+++ b/docs/option/language.aria.sortAscending.xml
@@ -10,14 +10,14 @@
 	<description>
 		ARIA label that is added to the table headers when the column may be sorted ascending by activating the column (click or return when focused).
 
-		Note that the column header text is prefixed to this string. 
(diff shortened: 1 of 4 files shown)


DataTables:
Think looks like a fantastic addition - thanks! I'm not yet sure if I'll include it in 1.10.13, or the following. Let me get back to you on that - I should be able to confirm next week.

DataTables:
Sorry for the delay in getting back to you here!

I feel inclined to include this in the next major update (be that 1.11 or 2.0) rather than in 1.10 and simply do away with the option to not have `_HEADER_` in the string. It would simplify the code a little and keep it compact.

Hopefully we aren't too far away from branching to the next major version, so I'll include this then.

PR #88, opened 2016-11-23
Update documentation to reflect that there are 6 built in paging options now, not 4

tziporaziegler:
After adding the first_last_numbers pagination option, I found some documentation that needs to be updated. I don't know if I found all the places, but this is definitely a start. As you can see, this wasn't even updated when the fifth option was added.


DataTables:
Oops - yes. That looks like a good change - thanks. I might generalise it "has the following built in options" in future!

PR #90, opened 2017-01-06
Force research when regex is active

piet23:
The problem is that when using regex and querying something like "item1|item2", only item1 is actually filtered, as the filter is always applied against the already filtered data. To ensure that the "or"ed query is properly applied, one possibility is to remove the last character and add it again (the "prevSearch.length > input.length" condition fires).

This fix forces a new search against the master array when regex is set to true, so that such "or"ed queries will behave as expected.

### DIFF
--- a/js/core/core.filter.js
+++ b/js/core/core.filter.js
@@ -240,6 +240,7 @@ function _fnFilter( settings, input, force, regex, smart, caseInsensitive )
 		// New search - start from the master array
 		if ( invalidated ||
 			 force ||
+			 regex ||
 			 prevSearch.length > input.length ||
 			 input.indexOf(prevSearch) !== 0 ||
 			 settings.bSorted // On resort, the display master needs to be


DataTables:
Thanks - this does seem like a sensible idea. I'd like to do some tests as it will make regex slower, but frankly I don't really see how that can be avoided unless you have any genius ideas?

piet23:
Unfortunately, I have no great ideas in that regard, no. For us the use of "or"ed queries is most important, so checking whether the added text contains a "|" might prevent the full research in many cases, but I assumed that this would not be general enough.

On a side note: One further problem when using regex we have stumbled over, and have not yet fixed, is the following: When regex is activated, whenever you enter something like a backslash, some paranthesis or other special characters, errors are logged in the console (from the transformation to a RegEx) until the search string can be successfully transformed into a RegEx again.

AllanJard:
v2+ always does a full search - no need to force any more. I forgot to update this one at the time, apologies. Closing now.

PR #93, opened 2017-01-19
Dev: Changed how the searchDelay works

piiih:
Hi, the "feature" that I added is about the searchDelay.
I changed the searchDelay to just call the search function after the user stops to type, instead of when the user starts to type.
I don't know if you'll like that ;D

Thanks.


DataTables:
Thanks for this PR - much appreciated.

I'm not going to pull it in right now, but it is something that I plan to do in the next major version of DataTables. When that branch is created I'll look at pulling it into there.

PR #99, opened 2017-03-20
Fix: Bootstrap 4 alpha 6 integration fixes for pagination

jvb187:
Updated the pagination container to include pagination-sm to match the rest of the controls.


DataTables:
Thank you!

PR #103, opened 2017-04-20
Update jquery.dataTables.scss

markvantilburg:
Remove old vendor prefix
-webkit-box-sizing


DataTables:
Yup - its probably about time to do that. Thanks for the PR.

PR #104, opened 2017-04-21
Footer scroll fix when using the bootstrap theme

smasala:
When using the bootstrap 3 theme and the option `scrollY: true`, the footer stops scrolling before the table does, causes the columns to be out of line with the footer.

**Example:**
https://jsfiddle.net/hctp0bg2/

**Fix:**
```
div.dataTables_scrollFoot  > .dataTables_scrollFootInner {
    box-sizing: content-box;
}
```

**On a side note:**
Whilst implementing this pull request it seems that the current selector `div.dataTables_scrollFoot > table { }` is never met. When `dataTables_scrollFoot` is present then the child `table` is always inside `dataTables_scrollFootInner` (not a direct child), the element with `border-top` are the `th` elements inside the `table ` and **not** the `table` itself. Hope this helps :+1:

### DIFF
--- a/css/dataTables.bootstrap.scss
+++ b/css/dataTables.bootstrap.scss
@@ -166,9 +166,15 @@ div.dataTables_scrollBody {
 	}
 }
 
-div.dataTables_scrollFoot > table {
-	margin-top: 0 !important;
-	border-top: none;
+div.dataTables_scrollFoot {
+	> table {
+		margin-top: 0 !important;
+		border-top: none;
+	}
+
+	> .dataTables_scrollFootInner {
+		box-sizing: content-box;
+	}
 }
 
 


DataTables:
Awesome thank you! Are you happy for this to be pulled in under the MIT license?

smasala:
:) welcome. Yea no prob!

DataTables:
Wish I could remember why I had that `>table` bit in there. It does seem utterly redundant! Thanks for pointing that out.

smasala:
how I feel everyday 👍  :) :)

When the y-scroll bar is there, then it makes sense to remove the margin and border. The `border-top` has since moved to the `th` elements though. However, if there is *no* scroll bar present, you need the border again :)
**Example:** https://jsfiddle.net/rbmhhote/

DataTables:
I've opened #105 to make sure I don't forget about it before the next release!

PR #109, opened 2017-09-15
Fix typo in `reverse()` documentation.

pradyumna2905:


DataTables:
Thanks for this! Are you able to make the change without changing the whitespace? It looks like the tabs have been converted to spaces.

PR #110, opened 2017-09-16
Avoid applying padding to non sortable items on small tables

os1r1s110:
I don't know if there are more to this or if it would break some other rules but as of my testing it should be fine to remove this rule for headers that have no sorting applied to them.

This is the case already for "normal sized" tables as the padding-right rule is only applied if there is a sorting class to the element. But it has been forgotten for the table-sm variant. So there are two options that I can see.

Specify this rule does not apply if there is a sorting_disabled class
Do the same as what was done for the normal sized table in the following commit: 4e611f31ff84deae5ae4666c7ba60ad8cb1012c3

### DIFF
--- a/css/dataTables.bootstrap4.scss
+++ b/css/dataTables.bootstrap4.scss
@@ -201,7 +201,7 @@ div.dataTables_scrollFoot {
 
 // Condensed
 table.dataTable.table-sm {
-	> thead > tr > th {
+	> thead > tr > th :not(.sorting_disabled){
 		padding-right: 20px;
 	}
 


DataTables:
Thanks for this!

The problem I've found in the past is that having different padding on different columns in the table can result in some weird things going on with column sizing when scrolling is enabled (since DataTables has to split the table into its component parts and then align the columns).

I can't find a commit for that now, so I might be imagining it, or it might have been somewhere else I found it, but I think this needs a little more investigation first.

os1r1s110:
No problem, this was just a suggestion as I made the modification in my local version to make it work with my app, but as you said, this might have some effects elsewhere that I might not know about or that might not affect me in my current use case.

DavidJohnWilliams:
Hi, I was about to create a PR for this myself as I noticed it in one of my applications. I agree with @os1r1s110 that this padding removal is done for normal sized tables. Also looking through the history it seems that the padding removal was added some time ago in the Bootstrap 3 css but wasn't copied to the table-condensed at the same time. This has subsequently carried over to the Bootstrap 4 css and then scss. I suspect that at the time the padding removal was added it would have been an error to not include in the table-condensed styles which was missed as it it not used as often.

The fix in this PR is sound, however I would suggest for consistency using the same rules as for standard table cells or swap them to use the :not as well.

DataTables:
Thanks for the PR. I'd forgotten about this one to be honest. I will look into it for the next patch release.

PR #113, opened 2017-11-21
Remove inline styles in injected HTML

### LINKED ISSUE #112: Inline styles injected by JavaScript blocked by Content-Security-Policy
[As discussed](https://www.datatables.net/forums/discussion/comment/107953), a `Content-Security-Policy` that blocks injection by denying inline styles will cause the browser to block DataTables' `_fnApplyToChildren` function, which violates that policy.

To test, send a CSP header in your HTTP response with `object-src 'none'` as a policy, like so:

    Content-Security-Policy: default-src 'self' https:; object-src 'none'

Observe CSP reports in the JavaScript console.

bitplane:
fixes #112

### DIFF
--- a/js/core/core.scrolling.js
+++ b/js/core/core.scrolling.js
@@ -340,14 +340,18 @@ function _fnScrollDraw ( settings )
 	// both match, but we want to hide it completely. We want to also fix their
 	// width to what they currently are
 	_fnApplyToChildren( function(nSizer, i) {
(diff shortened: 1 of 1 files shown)


DataTables:
Awesome! Thanks for the PR.

bitplane:
Absolutely, licensed as per [license-mit.txt](https://github.com/bitplane/DataTablesSrc/blob/remove-html-inline-css/license-mit.txt). :)

PR #116, opened 2018-02-07
Bug fix for table.columns(x).visible(false)

nullquery:
When calling `table.columns(x).visible(false)` the following error is thrown:

> TypeError: Cannot read property 'sWidth' of undefined

Apparently invisible columns are not taken into account when calculating the width of a column.

I have fixed the issue by adding a null check. The error no longer shows up in the JS console and the feature (hiding columns) works as expected.

### DIFF
--- a/js/core/core.scrolling.js
+++ b/js/core/core.scrolling.js
@@ -255,7 +255,7 @@ function _fnScrollDraw ( settings )
 
 	$.each( _fnGetUniqueThs( settings, headerCopy ), function ( i, el ) {
 		idx = _fnVisibleToColumnIndex( settings, i );
-		el.style.width = settings.aoColumns[idx].sWidth;
+		if (idx !== null) { el.style.width = settings.aoColumns[idx].sWidth; }
 	} );
 
 	if ( footer ) {


DataTables:
Thanks for the PR. Can you give me a link to a page showing the issue please? [This example](http://live.datatables.net/riyixifo/1/edit) doesn't result in that error.

nullquery:
Odd. I tested against the latest released version (1.10.16) and I was able to reproduce it there. Although I'm using both the FixedHeader and Scroller plugins (though Scroller is disabled - but plugin still loaded - on the table in question).

DataTables:
Adding Scroller and FixedHeader, enabling only FixedHeader [still doesn't result in any errors](http://live.datatables.net/keyufiqo/1/edit) I'm afraid. If you could provide a link to a page showing the issue that would be useful. I suspect that its a configuration issue and that error, if we add the null check, will just shunt the issue elsewhere or disguise it completely, but there would still be an error.

nullquery:
I can't provide a link because the page in question is only available on my local development environment. Code in production is behind a login. I can provide my configuration, maybe that will help?

`
table = $tableElement.DataTable({
	'columns': calculateColumns(),
	'serverSide': true,
	'autoWidth': false,
	'pageLength': options.pageLength || 10,
	'lengthChange': !options.scroller && options.lengthChange !== false,
	'ajax': {
		'url': (options.uriprefix || '') + '/queries/' + queryid + '/list',
		'type': 'POST',
		'data': function(d)
		{
			processAjaxData(d, false);
		}
	},
	'ordering': options.ordering,
	'order': options.orders,
	'language': JSON.parse($element.attr('data-language')),
	'paging': options.paging !== false,
	'pagingType': options.pagingType,
	'info': options.info !== false,
	'searching': options.searching !== false,
	'deferRender': true,
	'scroller': options.scroller ? { 'rowHeight': d.rowHeight - (isFirefox ? 1 : 0), 'displayBuffer': 10 } : false,
	'scrollY': options.scroller ? d.height : undefined
});
`

DataTables:
My guess is that `calculateColumns()` is resulting in something other than an array of the length that matches the columns in the HTML.

I'm afraid we'd need a test case on JSFiddle, CodePen, http://live.datatables.net or similar so we can add this into our own test cases before this could be pulled in. If you can create such a test case, that would be very useful.

nullquery:
In our use case we're adding a second row to the `thead` to display a search input per-column. This was not updated correctly.

The error occurred because `_fnGetUniqueThs()` looks at all `tr` elements in the `thead` rather than only the `tr[role=row]` element generated by DataTables. This was not clear from the thrown exception.

PR #117, opened 2018-02-10
destroy remove performance

ypconstante:
Currently DataTables does a lot of dom manipulations when destroying a table, to return the table to its original state, even when requested to remove the table from the dom. 

I changed the destroy method to just call the remove method on the wrapper element when removing the table.


DataTables:
I see the point there - its a good one. However, there is a risk of memory leaks with this approach. If event listeners have been applied inside the table to rows or columns which have been hidden, those event listeners won't be unbound automatically.

There is also the small edge case of having a hold of the `table` element in a variable reference, so you can use it later and you expect the table to properly be restored.

It could be that this is something which should just be noted in the documentation, but it would be a change in behaviour.

Issue #118, opened 2018-02-16
rows([ids], {search: 'removed'}) optimization


dagadbm:
Hello,

I am working on a project that has a lot of data in the datatables and there is a use case that I need to know which rows were removed. I decided , at first, to use the rows([mySelectedIds], {search: 'removed'}) filter but alas I had to do use search: 'filter' instead and remove the duplicates myself.

Why? Because datatables removed is O(n^m) since they do 2 chained fors.

I propose to convert the masterData array into a map and then just iterate the filteredData (this way you get O(n) instead. 

Please let me know what you think.

--- Off topic ---
I have never done a PR request and I have no idea how so I am not sure I am the best person to fix this but I just wanted to let you guys know about this possible alternative :)

I wouldn't mind learning how to do a PR but it seems I need to fork this repo always? Is this common practice among other project? Can someone give me some link to get started on understanding how open source contributions work?

Thanks again :)

David.

DataTables:
Hi David,

Thanks for opening this discussion - performance improvements are always welcome!

When selecting the rows that can be selected from, its [just a single `for` loop at the moment](https://github.com/DataTables/DataTablesSrc/blob/9df2759744d7d8fd50caac67b45b29545cc91fc6/js/api/api.selectors.js#L125) unless I'm missing something (highly likely!)?

For pull requests - yes, the process is that you fork the repo, make whatever changes you need, and then send a pull request so I can merge the changes in. [This page](http://www.firsttimersonly.com/) (and the links in it) contain information that might be of some interest.

dagadbm:
[I am talking about this line](https://github.com/DataTables/DataTablesSrc/blob/9df2759744d7d8fd50caac67b45b29545cc91fc6/js/api/api.selectors.js#L120-L121)  (which was the use case I had to use). 

Will read that links as well :) thanks!

DataTables:
Got it! $.inArray should be fast in a relatively modern browser since its a native call that jQuery makes (`Array.prototype.indexOf`). However, its still a for loop, just native in the Javascript engine, so it could be improved. An `if` condition there with a map conversion could I think make it O(2n), since you'd need an initial loop to do the conversion.

Certainly would be interested to see what you cook up there, and if you can provide a test case showing before / after, that would be fantastic.

dagadbm:
I will see what I can do. Depends on my time as well. If I need some help should I use this issue or do you have some different form of communication?

DataTables:
Here is good.

[...]

PR #119, opened 2018-02-16
rows([ids], {search: 'removed'}) optimization

dagadbm:
This is an optimisation that is very noticeable when using the rows([ids], {search: 'removed'}) in particular when the data table has a LOT of rows (like 50k+).

For more about this discussion check this [issue](https://github.com/DataTables/DataTablesSrc/issues/118).

### DIFF
--- a/js/api/api.selectors.js
+++ b/js/api/api.selectors.js
@@ -113,13 +113,22 @@ var _selector_row_indexes = function ( settings, opts )
 		}
 	}
 	else if ( order == 'current' || order == 'applied' ) {
-		a = search == 'none' ?
-			displayMaster.slice() :                      // no search
-			search == 'applied' ?
-				displayFiltered.slice() :                // applied search
-				$.map( displayMaster, function (el, i) { // removed search
-					return $.inArray( el, displayFiltered ) === -1 ? el : null;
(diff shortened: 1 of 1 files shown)


DataTables:
Awesome - thanks for this. It looks good to me.

One thing I would like to see changed is the first `if` and `if else` condition - they result in the same action so could be combined. I'm happy to make that change after pulling this in, or to take an updated PR?

dagadbm:
yes I have no problem with the licenses etc.

I'm not sure I understood what you meant by the if one is displayMaster slice other is displayFiltered slice so I will just let you do it . 

Also, I didn't do the "removed search" for the below else if (index/original) (I just noticed it) but I am not comfortable changing that because I don't really know how to test that case and I don't wanna break stuff.

DataTables:
I've moved the code around a little for that change and also to match the code styling of the rest of DataTables for consistency in 64b3e958bc86f3b2820e3a92e988c4463eb086e3.

I agree, something similar could be done for the index order. We are currently revamping the unit tests, and I'll make sure that this all gets covered in them.

dagadbm:
let me know if I can help in anything else :)

PR #120, opened 2018-02-19
New optional placeBefore parameter for row().child.show()

KarloX2:
Makes it possible to control the placement of the child rows, either immediately after the parent row (which is the default), or (as a new option) immediately before the parent row. This can be specified on a per-row basis.

### DIFF
--- a/docs/api/row().child.show().xml
+++ b/docs/api/row().child.show().xml
@@ -5,8 +5,13 @@
 	<since>1.10</since>
 
 	<type type="function">
-		<signature>row().child.show()</signature>
+		<signature>row().child.show( [placeBefore] )</signature>
 		<description>Show the child row(s) of a parent row</description>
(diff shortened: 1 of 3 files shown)


DataTables:
Excellent thank you for this! Are you happy for it to be included under the MIT license?

I might sit on it for a little while, before merging it in to see if there is any other interest in it, but that's a good patch - thanks!

KarloX2:
Thanks. Yes, MIT license is ok, of course.

KarloX2:
Just a brief explanation about how I use this feature:

A child row is placed before every regular data row with a message text (in a cell over the entire table width) and this message is the most important and leading piece of information for every record and is always visible. The cells/fields in the regular row beneath are logically details (like when-raised, level of importance, source of information etc.) of this message, and visibility of some of these columns can be selected, but the ("child-") message row can’t be hidden.

PR #128, opened 2018-08-07
Display a more relevant error when no column headers are present

mmcev106:
**Before:**
http://plnkr.co/edit/mov49plvZ1LcEXti1fsx?p=preview

Notice the fairly cryptic "Uncaught TypeError: Cannot read property..." javascript error.

**After:**
http://plnkr.co/edit/nOWFB4DjpYCdXkUdwpiF?p=preview

It was a hard decision, but I'm fine if these three lines get released under the MIT license.

### DIFF
--- a/js/core/core.constructor.js
+++ b/js/core/core.constructor.js
@@ -270,6 +270,11 @@ else
 	aoColumnsInit = oInit.aoColumns;
 }
 
+if ( aoColumnsInit.length === 0 )
+{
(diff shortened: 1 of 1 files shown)


DataTables:
Yup - good point. I'm actually working on a rewrite of the column handling at the moment, so I'm not going to pull it in right away, but I'll do so when I've got the rest of my tree sorted out.

PR #130, opened 2018-11-21
Add location.hash to local storage state key

jnpdev:
Append the current hash to the key of the local storage save state.


DataTables:
Thank you for this! I can see the benefit of this in SAPs, but equally, in some where the anchor is used for navigation it might be a little bit more confusing. Let me have a bit of a think about it!

PR #132, opened 2019-01-21
Add ColReorder dom option

Trainmaster:
See https://github.com/DataTables/ColReorder/blob/1.5.1/js/dataTables.colReorder.js#L1343


DataTables:
Yup - good call. Thanks. Happy to have this as MIT licensed (as trivial as it is, I should still ask :-) )?

Trainmaster:
@DataTables Sure.

PR #133, opened 2019-01-30
add _fnAjaxJsonSrc

efraimGit:
in order to enable use dataTables response as property of a server standart response


efraimGit:
hi,

i close this pull request because i want make it from another branch.

thanks

PR #134, opened 2019-01-30
add Ajax JsonSrc option

efraimGit:
hi,
i create this pull requst again from new branch to seperate every future to uniq branch.

explain:
in my project any request for rest api server answered with standart response like 
public class ApiResponce<T>
    {
        public HttpStatusCode Status { get; set; }
        public T ResponseContent { get; set; }
        public ResponseError Error { get; set; }
    }
to save this standard but use dataTables's expected response, i added this functionality,

to use it the user need only to added this row:
 "ajax": {
    "url": 'url'
     "jsonSrc": 'ResponseContent', // this row
  },


DataTables:
Is this not just what the `ajax.dataSrc` option in DataTables does already? https://datatables.net/reference/option/ajax.dataSrc

efraimGit:
No. With this option is possible use serverside processing with draw and records total and filtered properties.

DataTables:
Ah! I'm with you know. Thanks for the clarification. My plan is to address this with the `ajax.dataSrc` option, giving it the ability to support server-side processing as well (probably using an object doing some mapping).

I'll probably not pull this PR in I'm afraid as I think there could well be a bit of confusion about the parameters names (as my own proves), but this is something that will be addressed in future!

PR #135, opened 2019-02-13
Fix for table.ajax.reload() with data source "ssd" and ajax object in configuration

gregorschatz:
In Microsoft IE 11.0.40 and Apple Safari 12.0.3 I get additional
DataTable initialisations with data source detected as 'dom' and settings Object containing ajax set to null.
It nevertheless tries to build an ajax request and thus and breaks when
ajax from settions object comes into use.

var table = $("#table").DataTable({
	serverSide: true,
	ajax: {
		url:"...",
		method:"...",
		data: function (){
			// ...
		}
	}
});

$("#table").DataTable().ajax.reload();


DataTables:
Hi

Thanks for the PR. Are you able to give me a link to a test case showing the issue please? If you are also able to add a unit test for the issue it is addressing, that would be great and let me understand the problem more fully.

gregorschatz:
It will take some time but I try to construct a live example for the issue.

DataTables:
That would be excellent. Thanks.

Issue #139, opened 2019-05-06
jquery.dataTables.js throws an exception when rows have colspan (count does not match columns)


NateRadebaugh:
jquery.dataTables.js throws an exception when rows have colspan (count does not match columns)

Example codepen: https://codepen.io/NateRadebaugh/pen/BeyzVN?editors=1010

Look in the console and see:
![image](https://user-images.githubusercontent.com/130445/57232763-eb26ee80-6fe2-11e9-9954-1e1a4d2f3de4.png)

![image](https://user-images.githubusercontent.com/130445/57232804-07c32680-6fe3-11e9-9325-8bb164a03691.png)

Proposal:
![image](https://user-images.githubusercontent.com/130445/57232894-3c36e280-6fe3-11e9-9c43-2427460876d2.png)

NateRadebaugh:
By design.

PR #140, opened 2019-05-06
Stop processing columns if we've reached the end of tds with explicit columns

NateRadebaugh:
# Description
Stop processing columns if we've reached the end of `td`s with explicit columns

Resolves DataTables/DataTablesSrc#139

# Example
codepen of issue: https://codepen.io/NateRadebaugh/pen/BeyzVN?editors=1010

`jquery.dataTables.js` throws an exception when rows have `colspan` (`td` count does not match columns count)

## Look in the console and see:
image

## Existing code:
image

### DIFF
--- a/js/core/core.draw.js
+++ b/js/core/core.draw.js
@@ -39,6 +39,12 @@ function _fnCreateTr ( oSettings, iRow, nTrIn, anTds )
 			create = nTrIn ? false : true;
 
 			nTd = create ? document.createElement( oCol.sCellType ) : anTds[i];
+
+			// Stop processing columns if we've reached the end of tds with explicit columns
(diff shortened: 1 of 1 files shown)


DataTables:
Thank you for this! However, unfortunately its not quite that simple. colspan support in DataTables would need to be able to handle sorting over the multiple columns, hiding columns (column visibility) and search.

It would also present significant complications for API methods such as `cells()`, extensions such as data export, fixed columns and others.

I fear I'm not going to pull this in at the moment as it could be the beginning of a very large amount of work.

NateRadebaugh:
@DataTables I can understand that. If only designers understood technical limitations of libraries we don't control 😉

PR #152, opened 2019-11-19
Bugfix: Avoid "Uncaught TypeError: Cannot read property 'sWidth' of undefined"

takieszti91:
while using column hider function (for example datatables.net-buttons/colVis).

### DIFF
--- a/js/core/core.scrolling.js
+++ b/js/core/core.scrolling.js
@@ -129,7 +129,7 @@ function _fnFeatureHtmlTable ( settings )
 	}
 
 	$(scrollBody).css(
-		scrollY && scroll.bCollapse ? 'max-height' : 'height', 
+		scrollY && scroll.bCollapse ? 'max-height' : 'height',
 		scrollY
 	);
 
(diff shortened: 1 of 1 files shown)


DataTables:
Normally this issue is caused by defining more column than is in the HTML. Can you give me a link to an example showing the issue?

takieszti91:
I created 2 small example for this: https://github.com/takieszti91/ErrorExampleForDataTables
This error occurs when I use lightColumnFilter, scrollX and colvis at the same time.

takieszti91:
You have to hide column(s) with the button.

DataTables:
Thank you for the example. I believe it is caused by:

```
	for (var i = 0; i < 6; i++) {
		$scope.columnfilter_object[String(i)] = {html: 'input', type: 'text'};
	}
```

It is _always_ producing a 6 table column, but Angular is changing the DOM to only have 5 (or less) columns based on the click of the column selector. I believe the correct fix is actually to change that `6` to be dynamic based on the number of columns.

takieszti91:
I tried to change my code, but unfortunately I cannot change the original column filter. I can only add a new one. The error is still there. Can I change somehow the original column filter?

DataTables:
I'm afraid you'd need to ask the author of the column filter plug-in about that. That is not software that was developed by ourselves.

PR #154, opened 2020-01-13
Optionals messages for null values

LucasdaSilvaNi:
Hello everyone!

I would like to pull resquest what I believe it's useful new column definition. A property when developer doesn't want to show a blank cell for null values (I have worked with ajax)

For anything, can contact me.

Thanks, 
Lucas

### DIFF
--- a/js/core/core.data.js
+++ b/js/core/core.data.js
@@ -147,7 +147,7 @@ function _fnGetCellData( settings, rowIdx, colIdx, type )
 	}
 
 	if ( cellData === null && type == 'display' ) {
-		return '';
+        return settings.aoColumns[colIdx].nullValue;
 	}
 	return cellData;
 }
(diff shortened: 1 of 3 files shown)


LucasdaSilvaNi:
(Sorry, forgot a document which can be more clear)
[Funcionality explanation.docx](https://github.com/DataTables/DataTablesSrc/files/4051858/Funcionality.explanation.docx)

DataTables:
Thanks for the patch - but this is what the [`columns.defaultContent` parameter](https://datatables.net/reference/option/columns.defaultContent) already does. I'm not clear that this new option offers anything over `defaultContent`? Were you not aware of that option, or does it not do what you need?

LucasdaSilvaNi:
Well, what I supposed to pull is, in fact, what columns.defaultContent does already. So I'm closing this request.

Thanks for the reply

PR #156, opened 2020-02-20
Added code to deal with empty db and dt column

josiahking:
Modified pluck method to skip column with empty db and map correct array for filter to be effective
Modified data_output for better process of column with empty db
Modified filter to skip empty db column

### DIFF
--- a/examples/server_side/scripts/ssp.class.php
+++ b/examples/server_side/scripts/ssp.class.php
@@ -43,10 +43,20 @@ class SSP {
 
 				// Is there a formatter?
 				if ( isset( $column['formatter'] ) ) {
-					$row[ $column['dt'] ] = $column['formatter']( $data[$i][ $column['db'] ], $data[$i] );
+                    if(empty($column['db'])){
+                        $row[ $column['dt'] ] = $column['formatter']( $data[$i] );
+                    }
+                    else{
+                        $row[ $column['dt'] ] = $column['formatter']( $data[$i][ $column['db'] ], $data[$i] );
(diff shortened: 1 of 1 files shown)


DataTables:
Thanks for this. What's the intention with having an empty `db` parameter? I don't quite get the use case for that.

It looks like the whitespace has been changed to be spaces rather than tabs - or that might just be the github viewer?

And finally ( :-) ) if we pull this in, are you happy for it to be included under the MIT license?

josiahking:
Thanks for accepting this pull request in advance.
The use case is that you might need an action column where some buttons will function to delete or update a row, but the dt column is not a db column, so leaving the column db empty will cause php sql error.
For the empty dt column, you might need several db column that you might want to merge into one dt column using the formatter, so you will still need to select this db column, and use them later in a dt column.
E.g. 
```
array( 'db' => '',  'dt' => 'mid', 'formatter' => function($row){ return $row['rr']. " " .$row['rr_num']; }  ),
array( 'db' => 'surname',  'dt' => 'surname' ),
array( 'db' => 'first_name',  'dt' => 'first_name' ),
array( 'db' => 'email',  'dt' => 'email' ),
array( 'db' => 'rr',  'dt' => '' ),
array( 'db' => 'rr_num',  'dt' => '' ),
```

DataTables:
I'm with you now - thank you. Normally I've just done that kind f formatted on the client-side using a renderer, but I can see that if you didn't want the unformatted data to go to the client-side, then that would be useful.

PR #166, opened 2020-05-19
Fixes #165 - incorrect format for table-sm classes

### LINKED ISSUE #165: Wrong CSS selector for "table-sm" classes
For condensed tables (bootstrap class `table-sm`) there is a rule to add padding - to make space for the sort indicators.

See https://github.com/DataTables/DataTablesSrc/blob/1.10.21/css/dataTables.bootstrap4.scss#L210-L213

```
table.dataTable.table-sm {
	> thead > tr > th :not(.sorting_disabled){
		padding-right: 20px;
	}
```

I think this rule should be applied to the `<th>` element - not to all its children (and grandchildren).

i.e. you need to remove the space between `th` and `:not`.

My table headings contain a few nested spans and this rule is breaking them.  Removing the space fixes it.

fisharebest:
Fixes #165

I think this rule should apply to the `<th>`, and not to its children.

### DIFF
--- a/css/dataTables.bootstrap4.scss
+++ b/css/dataTables.bootstrap4.scss
@@ -208,7 +208,7 @@ div.dataTables_scrollFoot {
 
 // Condensed
 table.dataTable.table-sm {
-	> thead > tr > th :not(.sorting_disabled){
+	> thead > tr > th:not(.sorting_disabled){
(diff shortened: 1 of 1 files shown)


DataTables:
Thank you! Agreed.

PR #167, opened 2020-06-01
num-fmt does not sort correctly with French locale.

ferrarimartin:
In French, thousands are separated by a space. Space is not considered as thousands separator by `num-fmt`, and as a result, sorting of numbers is incorrect.

I've added regular space, and also non-breakable space ('\u00A0') as possible thousands separators, as .net core adds a non-breakable space instead of a regular one when calling `decimal.ToString()`.

This sample code shows the problem. The "Salary" column sort is incorrect: https://jsfiddle.net/ferrarimartin/oxLzrjgn/1/.

### DIFF
--- a/js/core/core.internal.js
+++ b/js/core/core.internal.js
@@ -43,7 +43,7 @@ var _re_escape_regex = new RegExp( '(\\' + [ '/', '.', '*', '+', '?', '|', '(', 
 // - Ƀ - Bitcoin
 // - Ξ - Ethereum
 //   standards as thousands separators.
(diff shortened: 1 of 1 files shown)


DataTables:
Thank you! I'll definitely take in the `'\u00A0` change, but I'm not sure about the regular space character. We support narrow space, for thousands and also comma or period as a thousands separator.

My consider with a regular space is that it is not uncommon to have a list of numbers which are space separated and it wouldn't make sense to order them as if it were a single large number.

If you could update the PR with that, that would be great - thanks!

ferrarimartin:
Thank you! Sorry for not thinking about the regular space other use cases. I've updated the branch, let me know if it's ok like this.

Thanks again,
     Martín.

DataTables:
Perfect - thank you.

PR #168, opened 2020-06-08
stripHTML to remove any HTML wrappers on the data prior to cell data being set

PeterCassell92:
I am using a platform called Verj.io to create my web application. Verj.io creates tables with excess HTML within the <td> for each cell in the table. This is part of the way that Verj.io builds controls.

I found that when using datatables, some functionality (such as SearchPanes) would not work correctly because the HTML within the <td> caused each entry to be unique so every entry had a frequency of 1. Also, the searchPane ended up misforming its own rows as the HTML would be escaped incorrectly.

I have added an option to allow people using a platform like Verj.io to treat the tables as if the data were directly in the <td> element. I have allowed for this by using a stripHTML flag that can be defined in the ColumnDefs. This means that the HTML generated by Verj.io is not included in the cell data.

### DIFF
--- a/js/core/core.data.js
+++ b/js/core/core.data.js
@@ -618,7 +618,11 @@ function _fnGetRowElements( settings, row, colIdx, d )
 	var cellProcess = function ( cell ) {
 		if ( colIdx === undefined || colIdx === i ) {
 			col = columns[i];
-			contents = $.trim(cell.innerHTML);
+			//if specified, read data from last child of cell
+			if(col.srcDataFromLastChild && cell.lastChild){
+				cell = cell.lastChild;
+			}
+			contents = (cell.innerHTML)? cell.innerHTML.trim(): '';
 
 			if ( col && col._bAttrSrc ) {
 				var setter = _fnSetObjectDataFn( col.mData._ );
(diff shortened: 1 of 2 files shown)


DataTables:
Hi,

Thanks for the PR!

Could you show me an example of the markup created by Verj.io please? Does it have a wrapper span tag or something?

I'm not sure about pulling this in to be honest, as its the first request we've had for such a thing, but if it is popular enough, then we can certainly consider it.

PeterCassell92:
VerjioTD
VerjioTD2

The Verj.io table markup may come with 1 or 2 Div wrappers.

The problem that this causes with DataTables is twofold:-
**1) searchPanes do not work correctly**

datatablesBroken
verjSearchPaneHTML

**2) it it inefficient to process so many extra characters when the data can be extracted more directly**

So this is the problem I'm trying to solve. I think that my updated solution does this fairly succinctly but if there is another way to do this already in DataTables then let me know. Alternatively, this kind of functionality could be something that's built into the 'data' attribute rather than creating a new attribute like I have.

DataTables:
Thanks for the screenshots. I agree - your solution to this is a nice one. My consider is that it is extra DOM checks (and thus slow) and extra code for one specific case. I also think that if you were to call `row().invalidate('data')` on your table, it would destroy the old HTML, thus possibly breaking some other interactions.

If it is a popular request, then I am happy to include it (or possibly we could make the DOM reading a plug-in system)

We'll look at what we can do with this with SearchPanes.

PR #172, opened 2020-06-22
Closes #171

### LINKED ISSUE #171: jQuery.isArray is deprecated; use Array.isArray
Quite a few `$.isArray` usages in the code that need switching to `Array.isArray`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

Not sure if the polyfill is needed, see above for browser compatibility.

bytestream:
Replaced deprecated `$.isArray` usage with `Array.isArray` as advised @ https://api.jquery.com/jQuery.isArray/

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray for browser compatibility.

### DIFF
--- a/js/core/core.internal.js
+++ b/js/core/core.internal.js
@@ -271,3 +271,10 @@ var _unique = function ( src )
 	return out;
 };
 
+// Array.isArray polyfill.
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray
+if (! Array.isArray) {
+    Array.isArray = function(arg) {
+        return Object.prototype.toString.call(arg) === '[object Array]';
+    };
+}
--- a/old-tests/tests_onhold/4_server-side/ajax.js
+++ b/old-tests/tests_onhold/4_server-side/ajax.js
@@ -532,7 +532,7 @@ $(document).ready( function () {
 		},
 		function () {
 			console.log( result );
-			return $.isArray( result[0] ) && result[0].length === 35;
+			return Array.isArray( result[0] ) && result[0].length === 35;
 		}
 	);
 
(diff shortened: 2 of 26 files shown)


js/core/core.internal.js
```diff
@@ -271,3 +271,10 @@ var _unique = function ( src )
 	return out;
 };
 
+// Array.isArray polyfill.
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray
+if (! Array.isArray) {
+    Array.isArray = function(arg) {
```
bytestream, line 277: Leaks into global space (if unset), is that OK? The rest of the functions in here are localised

old-tests/tests_onhold/4_server-side/ajax.js
```diff
@@ -532,7 +532,7 @@ $(document).ready( function () {
 		},
 		function () {
 			console.log( result );
-			return $.isArray( result[0] ) && result[0].length === 35;
+			return Array.isArray( result[0] ) && result[0].length === 35;
```
bytestream, line 535: Not sure if polyfill from `core.internal.js` is available in test files, or if its required?

DataTables:
Excellent - thank you for this. I think it is okay for the polyfill to drop into global space. I can't really see why that would be a problem unless someone is using that property to test for older browsers rather than using something more appropriate.

DataTables:
I forgot to ask - are you happy with this being released under the MIT license?

bytestream:
> 
> 
> I forgot to ask - are you happy with this being released under the MIT license?

Bit late now ;) but yes no worries haha.

DataTables:
I could remove it if you didn't. But that you for confirming that :-)

PR #175, opened 2020-07-06
Remove deprecated jQuery.trim()

taskinoz:
`jQuery.trim()` has been deprecated as of jQuery 3.5

[jQuery.trim() Doc](https://api.jquery.com/jQuery.trim/)

I've replaced the jQuery trims with `String.prototype.trim` to keep it compatible.

### DIFF
--- a/examples/resources/demo.js
+++ b/examples/resources/demo.js
@@ -22,14 +22,14 @@ if ( window.$ ) {
 
 		// css
 		var cssContainer = $('div.tabs div.css');
-		if ( $.trim( cssContainer.find('code').text() ) === '' ) {
+		if ( ( cssContainer.find('code').text() ).trim() === '' ) {
 			cssContainer.find('code, p:eq(0), div').css('display', 'none');
 		}
 
 		// init html
 		var table = $('<p/>').append( $('table').clone() ).html();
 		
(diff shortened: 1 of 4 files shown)


DataTables:
Thanks for this. DataTables core (1.x) supports back to IE6 though, so we'll need to add a polyfill for trim.

taskinoz:
There is a polyfill available on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
 
```js
if (!String.prototype.trim) {
  String.prototype.trim = function () {
    return this.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '');
  };
}
```

taskinoz:
I could go ahead and add the MDN polyfill to the current pull request so its all in one merge.

Would you also need to add it to the API references?

DataTables:
If you could put the polyfill into `js/core/core.internal.js` that would be great - thanks.

No need to add that to the documentation since it wouldn't be part of our official API.

taskinoz:
The polyfill has now be added to `js/core/core.internal.js`

DataTables:
Thanks! Are you happy for your changes to be included under the MIT license?

taskinoz:
Yes

PR #178, opened 2020-08-13
Replacing icons by DataURIs

Clorr:
The icons are really small, but as external files, they hit the network which is useless for such small resources. Ideal solution would be to generate DataURIs at compile time, but for now, this may be a step toward improvment.


DataTables:
Thanks! We are going to replace the icons with utf8 characters in v2 (actually, that is already done in the 2 branch). Our next release is going to be a patch (1.10.22) so it probably isn't quite the right thing to do for a patch, but I fully agree that this is a welcome step in the right direction.

Clorr:
Using chars seems the perfect solution ;-)

PR #181, opened 2020-10-27
Remove workaround for old WebKit/Blink

alexhenrie:
The referenced bugs were fixed over 5 years ago:

https://bugs.webkit.org/show_bug.cgi?id=131819
https://bugs.chromium.org/p/chromium/issues/detail?id=365619


DataTables:
Thank you. I agree that it is time to remove the workaround now.

alexhenrie:
Thank you!

PR #182, opened 2020-10-28
Remove redundant check from _fnCreateTr

alexhenrie:
In this function `create == !nTrIn`, always, so checking only one of the two is fine.


DataTables:
Goodness me - how did you spot that! Yup I agree.

Are you happy for this to be included (or removed in this case) under the MIT license?

alexhenrie:
Yes, of course all of my contributions to DataTables may be used under the MIT license at https://github.com/DataTables/DataTablesSrc/blob/master/license.txt

PR #183, opened 2020-11-21
Fix - Example (api)

SAPMarco:
Adjusted the wording on the example title. Looked like something was missing.


SAPMarco:
After reading over it a few more times ... it started making sense. I got confused, my bad.

DataTables:
No worries - thanks for the update. Is there anything we can do to make the doc clearer from the outset?

SAPMarco:
Not as of now. I haven't had any issues when looking for anything within the documentation but this once. So as far as I'm concerned, it's all good! =)

PR #186, opened 2021-01-08
Update dataTables.jqueryui.js

lukasz-schab:
Fixing duplicated sort wrapper.


DataTables:
Thanks for this! Are you happy for the change to be included under the MIT license?

Also, under what conditions does this happen? Is it a destroy which doesn't unwrap the header and then you need to initialise it again?

lukasz-schab:
Off course MIT it's not a problem.. It happened when ussing jQuery UI
Accordeon included in dialog box. After reopening the dialog box datatable
header is creating next sort icon. IT happened after upgrafing from 1.9 to
1.10

Pt., 8 sty 2021, 18:24 użytkownik Allan Jardine <notifications@github.com>
napisał:

PR #192, opened 2021-05-10
- Fixed issue with concurrent ajax draws being ignored and only rende…

skylord123:
…ring the first.

- Removed `settings.bAjaxDataGet` variable (now unnecessary as this needs to be tracked per each ajax request not globally)

I'm going to copy the description from https://github.com/DataTables/DataTables/pull/1083 since I accidentally create that PR before this correct one:

I was having issues when my users would fill out the filter column fields at the top of a table I have it set to filter on the change event of the text fields. When they would enter one value in a filter field, tab to the next value, enter a search value and press enter before the first request finishes (our tables are pretty large so requests take a few seconds). it would only filter what the user put in the first field and would drop subsequent requests until the first request finishes.

I traced it down to these lines of code. If multiple ajax requests happen the _fnDraw will drop the subsequent requests. This is because the _fnDraw function calls _fnAjaxUpdate to get the ajax data then once that data is available it calls _fnDraw again and this time it renders the content because settings.bAjaxDataGet is true.

Instead of storing the ajax request state in a variable on the datatable object it should instead only be used for that single ajax request so it doesn't break concurrency.

### DIFF
--- a/js/core/core.draw.js
+++ b/js/core/core.draw.js
@@ -363,8 +363,9 @@ function _fnDraw( oSettings, ajaxComplete )
 	{
 		oSettings.iDraw++;
 	}
-	else if ( !oSettings.bDestroying && !ajaxComplete && !_fnAjaxUpdate( oSettings ) )
+	else if ( !oSettings.bDestroying && !ajaxComplete)
 	{
+		_fnAjaxUpdate( oSettings );
 		return;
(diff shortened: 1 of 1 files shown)


DataTables:
This looks really good to me - thanks. When I merge it in it will go through our unit tests.

PR #197, opened 2021-12-18
Correct typo

kainung:


AllanJard:
Many thanks for this, but of you look at the code, you'll see that the `async` option is set to false, making it a synchronous request. Therefore Sjax is the correct acronym in this case - Synchronous JavaScript and XML.

PR #198, opened 2022-01-23
Fix to not always expect return string in _fnGetCellData to insert in innerHTML; and permit to call a callback 

amnambiar:
Hi Team,

IMPORTANT

Please test and accept this PR which will faciliate the following:
```
col.renderReturnType = 'function'; // only then will these changes be activated
const createdCellCallback = (tdElm, cellData, rowData, rowIndex) => {
  // dynamically insert an angular component into tdElm with help of Renderer2; passing in needed values from rowData and cellData as @Inputs to the component
}

col.render = (data, type, row, meta) => { return createdCellCallback; }

col.createdCell = createCellCallback;
```

This will faciliate blindly performing the function within returned function of col.render at all times (especially after reorder of columns)

This is in accordance to my communication over the issue thread https://github.com/DataTables/ColReorder/issues/49

Cc: @AllanJard, @SandyDatatables @colin0117  Please look into this. I would like this feature included in matter of hours to be exact, as my app relies on this! Please HELP

### DIFF
--- a/js/core/core.data.js
+++ b/js/core/core.data.js
@@ -150,7 +150,7 @@ function _fnGetCellData( settings, rowIdx, colIdx, type )
 	else if ( typeof cellData === 'function' ) {
 		// If the data source is a function, then we run it and use the return,
 		// executing in the scope of the data object (for instances)
-		return cellData.call( rowData );
+		return cellData;
 	}
 
 	if ( cellData === null && type === 'display' ) {
@@ -311,7 +311,13 @@ function _fnInvalidate( settings, rowIdx, src, colIdx )
 			cell.removeChild( cell.firstChild );
 		}
 
-		cell.innerHTML = _fnGetCellData( settings, rowIdx, col, 'display' );
+		var cellData = _fnGetCellData( settings, rowIdx, col, 'display' );
+		if (typeof cellData === 'string') {
+			cell.innerHTML = cellData;
+		} else if (typeof cellData === 'function') {
+			var targetCol = settings.aoColumns[col]
+			cellData(cell, targetCol.mData ? row._aData[targetCol.mData] : col.sDefaultContent, row._aData, rowIdx)
+		}
 	};
 
 	// Are we reading last data from DOM or the data object?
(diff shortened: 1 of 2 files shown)


anu91ab:
Can you please merge this PR in immediate effect in a matter of less than 24 hours, considering this a priority. PLEASEEE!

hhariharan94:
+1

AnuShankar93:
+1

AllanJard:
Many thanks for the PR. However, I'm afraid I won't be able to pull it in yet for the following reasons:

1. No example showing the changes in use
2. No documentation updates relating to the changes (it looks like there is a new `renderReturnType` parameter which isn't defined anywhere
3. I don't actually fully understand the change. Why can't you just pass the extra information into the `cellData` function rather than using a switch to call it with a different signature?
4. I'd like to see an example showing how you handle orthogonal data types such as sorting data as well as a simple example. It isn't clear to me how that is handled with these changes.

>  I would like this feature included in matter of hours to be exact, as my app relies on this! 

I'm afraid it doesn't work like that. I don't place demands on your time - I would appreciate if you didn't make demands of ours.

anu91ab:
@AllanJard ,

 http://live.datatables.net/qonidogo/6/ here is the link to reproduce the issue.

Reorder the columns and see that the links in Name column disapper.

I want the table to behave in a way that if , commented lines are uncommented in place of existing 'render', then links be inserted no matter what.

Feel free to suggest alternative key values for renderReturnType, and I can update the PR accordingly. Please let me know what else I can do to get this in!

anu91ab:
Regarding your point 3 above, I do not want the cellData to return a string back to be inserted into the innerHTML. Instead I want the entire insserting of data to be handled within the cellData function itself.

anu91ab:
Regarding 2, Yes it is a new Key i introduced to faciliate deciding if cellData needs to be invoked then and there using .call or if the cellData needs to be considered as a function to be returned, to be invoked when cell html is to be placed.

anu91ab:
I have raised https://datatables.net/forums/discussion/71345/customize-render-callback-to-not-expect-a-string-and-just-perform-actions-within-the-callback-fn  as well in place. Please see what can be done from your side.

Issue #200, opened 2022-03-09
Bootstrap 5 table-striped corrupted by dataTables.bootstrap5.scss when rowGroup is used


programatix:
I'm using the following,
```
<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/v/bs5/dt-1.11.5/r-2.2.9/rg-1.1.4/datatables.min.css"/>
<script type="text/javascript" src="https://cdn.datatables.net/v/bs5/dt-1.11.5/r-2.2.9/rg-1.1.4/datatables.min.js"></script>
```

When I enabled rowGroup, the table-striped is corrupted as such,
![image](https://user-images.githubusercontent.com/3293865/157422224-7f939ee8-1142-4639-926b-64fa2b851a8c.png)

I noted that all effected even rows has the following,
![image](https://user-images.githubusercontent.com/3293865/157421318-5848a7dc-8bda-4a89-9325-f7902ca654df.png)

I tried linking the css individually and if I removed dataTables.bootstrap5.css, then table-striped works fine even with rowGroup (both its css and js).

AllanJard:
It [appears to work okay in this example](https://datatables.net/extensions/rowgroup/examples/styling/bootstrap5.html).

Perhaps you could open a support thread [in the forums](//datatables.net/forums) with a link to a test case showing the issue please?

PR #201, opened 2022-03-10
Bootstrap v5.1.2 and above table-striped compatiblity

programatix:
This is the fix for Bootstrap v5.1.2 and above. I've reported this issue at https://github.com/DataTables/DataTablesSrc/issues/200


AllanJard:
Super - many thanks for this. Just to check though, this change will then not work with Bootstrap <=5.1.1. Or possibly it will work, but operate on the cell background rather than the row which 5.1.1 is presumably using (I haven't checked yet - will do so soon).

programatix:
I did a quick test with Bootstrap 5.1.0 and 5.1.1 and it works.

AllanJard:
Awesome - I've just checked as well and it looks good to me. Thanks for this :)

PR #204, opened 2022-04-01
Replaced outdated IE7 CSS hacks

eldang:
Parcel versions 2.4.0 & 2.4.1 refuse to build if any dependency has invalid CSS.  This has created an incompatibility with various other projects including DataTables, and the Parcels maintainer has expressed a strong preference for fixing all the other projects rather than loosening the constraint: https://github.com/parcel-bundler/parcel/issues/7854

Fortunately, in this project the issue seems to be confined to a handful of CSS declarations that use the very nonstandard [*propertyname hack for IE7 and older](https://stackoverflow.com/questions/1690642/purpose-of-asterisk-before-a-css-property).  This PR simply replaces each instance of that hack with the modern, standards-compliant [`@supports` construct](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports).  In each instance, the IE-specific rule is first, followed by an `@supports` block that overrides it for more modern browsers.  It's written in this slightly counterintuitive way because IE does not itself support `@supports`.

I would also be happy with dropping IE<=7 support entirely, or allowing a little degradation in those very old browsers by dropping these specific CSS rules.  But I do not feel that that's my decision to make!  So this is an attempt to maintain that support while complying with modern CSS standards.

Unfortunately I am not able to test this work in IE as I work on a Mac.

All work in this PR is offered under the MIT License.

### DIFF
--- a/css/dataTables.jqueryui.scss
+++ b/css/dataTables.jqueryui.scss
@@ -48,11 +48,9 @@ table.dataTable {
 			margin-left: 2px;
 			text-align: center;
 			text-decoration: none !important;
+			cursor: hand;
 			@supports (cursor: pointer) {
 				cursor: pointer;
-			}
-			@supports not (cursor: point) {
-				cursor: hand;
 			}
 
 			border: 1px solid transparent;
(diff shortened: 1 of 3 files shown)


eldang:
In case this comes up for anyone else: one of the Parcel maintainers suggested some workarounds at https://github.com/parcel-bundler/parcel/issues/7854#issuecomment-1116745050

I just tried https://github.com/ds300/patch-package and it was easy to apply and solved the problem for me.

PR #205, opened 2022-05-13
Added ssp_mysqli.class.php as a mysqli alternative to PDO ssp.class.php

Beat2er:
Hello, 
as I needed the ability to use the PHP ssp script, but I am using the mysqli driver, I modified (a copy) of the [ssp.class.php](https://github.com/DataTables/DataTablesSrc/blob/e1ec0e3b1364a14ead34aaaadeae6b13df9790f8/examples/server_side/scripts/ssp.class.php) to address that issue.
Cheers

### DIFF
--- a/examples/server_side/scripts/ssp_mysqli.class.php
+++ b/examples/server_side/scripts/ssp_mysqli.class.php
@@ -1,0 +1,590 @@
+<?php
+
+/*
+ * Helper functions for building a DataTables server-side processing SQL query
+ * In contrary to ssp.class.php this one uses the MySQLi driver and not PDO
+ *
+ * The static functions in this class are just helper functions to help build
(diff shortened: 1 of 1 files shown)


AllanJard:
This is awesome - thanks for sharing this. I'm not going to merge it in I'm afraid though, as it isn't something we want to fully support at this time. The existing SSP class is really just a demo of how server-side processing can be done. We tried doing [a bit of a gallery of different SSP scripts before](http://legacy.datatables.net/development/server-side/) but it quickly became unmaintainable.

What you might consider doing is posting it in [our forum](//datatables.net/forums) so others can find and use it.

Beat2er:
Ok, at least if someone is desperate, he can find this PR and take the code ;)

PR #209, opened 2022-06-06
Fix: Searches for special characters

BerkinAKKAYA:
We had a problem searching for characters like "Ğ, Ş, İ". I am not sure if it's the correct place to add support for these characters but we fixed it with this patch in our project.


AllanJard:
[This plug-in](https://datatables.net/plug-ins/filtering/type-based/diacritics-neutralise) is probably what you want for that.

PR #211, opened 2022-06-18
Updating title

salim114:
I think it's about Auto-locale display (Moment.js)


AllanJard:
Oops - thanks for catching that!

PR #214, opened 2022-07-08
functions and options missing from types but seem to be still present/available

hchiam:
Hi! I'm trying to upgrade DataTables from v1.10 up to v1.12.1 for a private project, but it seems that a few different functions and options were taken away in the updates as early as v1.11, but when I run the project I can still see these options are available in the console log in dev tools. Are the functions in the diffs for this PR all coming from https://github.com/DataTables/Select/blob/master/types.d.ts ? and the additions to `interface Settings` at the bottom of the diffs in this PR?

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -140,6 +140,16 @@ declare namespace DataTables {
         //#endregion "Row/Rows"
 
         /**
+         * Selects rows.
+         */
+        select(): void;
+
+        /**
+         * Deselects rows.
+         */
+        deselect(): void;
+
+        /**
          * Plug-in to sum the values in a data set.
          * 
          * @returns number representing the sum of the data set.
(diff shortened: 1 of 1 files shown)


hchiam:
Below are a couple demos. The code itself works, but the private project can't build because the types failed, which makes me think that the types for https://github.com/DataTables/Select didn't get installed properly for some reason. 
- DataTables 1.10.23: https://codepen.io/hchiam/pen/jOzqMJm 
- DataTables 1.12.1: https://codepen.io/hchiam/pen/poLyEYr

hchiam:
oh https://github.com/DefinitelyTyped/DefinitelyTyped/commit/3e1d0b037f49c431aafec0f3a4ad4184c0c45471 i will have to figure out why `.select()` doesn't show up for me on intellisense, even though `select()` is clearly in the `interface RowsMethods` for `@types/datatables.net@1.10.22`

PR #215, opened 2022-07-08
Update types.d.ts

Avaruz:
Takeit From https://datatables.net/reference/option/createdRow

Since 1.10.17: The cells for the column.


AllanJard:
Thanks!

PR #216, opened 2022-07-26
Fix types.d.ts

JMolenkamp:
Add missing colon between argument and type in types.d.ts


AllanJard:
Thanks for spotting that and the PR!

PR #221, opened 2022-09-13
Fix `js/integration/dataTables.bootstrap5.js`

WinterSilence:
- Add button's column to `dom`
- Keep `ul.pagination` attributes added by events (for example `initComplete`)


AllanJard:
Thanks for this - however, Buttons is not a core component of DataTables. It is an optional extra so it wouldn't be appropriate to include it in the core DataTables files. As such, I'm going to close this PR. Happy to talk further about it if you like though.

WinterSilence:
@AllanJard ok, send second fix in #222

PR #230, opened 2023-01-26
prevent lengthMenu manipulation through the dom

DrAlghazali:
this will add a way to prevent a server hit on large tables if the end user has changed the length value through the dom

lengthMenu

### DIFF
--- a/js/core/core.length.js
+++ b/js/core/core.length.js
@@ -1,8 +1,15 @@
 
 function _fnLengthChange ( settings, val )
 {
-	var len = parseInt( val, 10 );
-	settings._iDisplayLength = len;
+	// val: is coming directily from the dom 
+	var len = parseInt(val, 10),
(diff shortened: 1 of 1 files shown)


AllanJard:
Thanks for the PR. I've wondered about this sort of thing before, but never implemented it as it is still trivial to bypass. Just intercepting the XHR request and then modifying the value, or changing the initialisation objects of DataTables. As a result, I'm inclined not to pull this in because it is still easy to bypass, and thus just adds a little extra weight to the code base.

Is this a problem you've run into - people requesting more data than they should? The only "real" fix is to make use that the page length is 10 / 25 / 50 / whatever at the server-side and enforcing it there.

DrAlghazali:
thank you Allan, yes the server the side is main way to go with this kind of issues. it was meant to make a quick fix for such an issue from the core client side. I already have implemented a server side fix for this on my Editor code

PR #231, opened 2023-01-28
types: Fix column definitions and oSort

apeschar:
Two fixes for the TypeScript definitions:

1. Require either `targets` or `target` on a column definition. The current types don't work because they require `target` to be specified, even when `targets` are given.
2. Add `oSort` on the definitions for `ext`.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -521,12 +521,16 @@ export interface ConfigColumns {
     width?: string;
 }
 
-export interface ConfigColumnDefs extends ConfigColumns {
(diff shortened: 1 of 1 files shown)


AllanJard:
I like it - thanks for the PR. Are you happy for this to be merged under the MIT license?

apeschar:
> I like it - thanks for the PR.

Glad to hear. :)

> Are you happy for this to be merged under the MIT license?

Yes.

AllanJard:
Nice one - thank you :)

PR #240, opened 2023-07-13
Use aria-label from table header if available

andreblanke:
I was surprised to find that `aria-label` attributes of `th` elements were getting overwritten by DataTables even though I did not use the `columns.ariaTitle` option. The overwrite is caused within `_fnSortAria` which assigns either the value of the `columns.ariaTitle` option or the column title to `sTitle`:

```js
var sTitle = col.ariaTitle || col.sTitle.replace( /<.*?>/g, "" );
```

In my opinion, an explicitly specified `aria-label` attribute should take precedence over deducing a label from the column title via `col.sTitle.replace( /<.*?>/g, "" );` which would then result in the following priorities:

1. A `columns.ariaTitle` value provided within the script
2. An existing `aria-label` provided within the document
3. A generated label derived from the column title as a fallback

## Context

For a bit of additional context: I am templating HTML on the server. For localization purposes it would be easiest to directly write a localized `aria-label` into the generated document.  
However, because the attribute currently gets overwritten by DataTables, the label text must be stored elsewhere (e.g., a `data-*` attribute or a hidden `div`), loaded from DOM, and then assigned to `columns.ariaTitle`, which feels a bit cumbersome.

I'd be glad to have my contribution included under the MIT license.

### DIFF
--- a/docs/option/columns.ariaTitle.xml
+++ b/docs/option/columns.ariaTitle.xml
@@ -10,7 +10,7 @@
 	</default>
 
 	<description>
-		This option can be used to provide a custom string for the `aria-label` attribute used for column headers to provide enhanced accessibility for screenreaders. By default DataTables will use the column title (`-init columns.title`, or read from the document), but this option provides customisation for that.
+		This option can be used to provide a custom string for the `aria-label` attribute used for column headers to provide enhanced accessibility for screenreaders. By default, DataTables will use an existing `aria-label` attribute or the column title (`-init columns.title`, or read from the document), but this option provides customisation for that.
 
 		Please note that the value given here will have `-init language.aria.sortAscending` or `-init language.aria.sortDescending` appended to it if the column can be ordered by the end user, to give an indication to screenreader users that this is the case. 
 	</description>
--- a/docs/option/columns.ariaTitle.xml
+++ b/docs/option/columns.ariaTitle.xml
@@ -10,9 +10,9 @@
 	</default>
 
 	<description>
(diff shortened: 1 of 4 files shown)


AllanJard:
Excellent point. I'm a bit surprised we've not encountered this use case before as well, now that you mention it!

The `_fnSortAria` function can be called multiple times (once per draw), so I don't think the change or reading the attribute will work, since it would just read back what it has already written on a previous draw. I think in the column's initialisation it would need a check to see if there is already an `aria-label` attribute, and if so cache that for use in `_fnSortAria`.

Would you like to look into doing that, or shall I?

andreblanke:
Ah, I see. Sorry, I mistakenly thought that `_fnSortAria` would get called only once at initialization time, but of course it is once per draw.

> Would you like to look into doing that, or shall I?

I'd like to look into it first. Thank you for your hints!

andreblanke:
That should hopefully do it. It might be worth to consider moving `_fnSortAria`'s `col.sTitle.replace( /<.*?>/g, "" )` to the new initialization code within `_fnColumnOptions` so that everything is in one place, but for now I wanted to make as few changes as necessary:

```javascript
// Fall back to the aria-label attribute on the table header if no ariaTitle is
// provided.
oCol.ariaTitle ||= th.attr("aria-label") || oCol.sTitle.replace( /<.*?>/g, "" );
```

Please let me know what you think about it. If it looks fine I'd squash the commits (and maybe move the `col.sTitle.replace( /<.*?>/g, "" )`).

AllanJard:
Possibly the best pull request I've had! Thank you :).

andreblanke:
Thank you, both for the positive feedback and the quick merge. It was a pleasure working with you. :)

PR #243, opened 2023-09-08
Fix: Sorting Icon Position of Bootstrap 5 table-sm tables

blankse:
The right padding is reduced from 26px to 20px. But the position of the icons is still 10px from right. The icons have a width of 10px. So the text have no spacing to the icons:
Screenshot 2023-09-08 100130

After my change the icons are 5px from right. So it have also 5px spacing to the text:
Screenshot 2023-09-08 100145


AllanJard:
Yup - good change - thank you! [This example](https://live.datatables.net/firogize/1/edit) shows the extra padding which should be reduced down to 5px.

blankse:
@AllanJard Yes, I am happy for this to be included under the MIT license.

PR #247, opened 2023-11-01
http to https

vanillajonathan:
Change links from http to https.


AllanJard:
Many thanks for this - looks like a good change set.

PR #249, opened 2023-11-14
[BUG] fix DT overwriting children td colspan when td already has a colspan.

murilo-plantae:
Consider a table that has 7 columns, and the following child template:

```html
<template>
    <tr>
        <td>Something...</td>
        <td colspan="5">Something...</td>
        <td>Something...</td>
    </tr>
</template>
```

When DataTables render it, the colspan attribute get overwritten by the visible column count breaking the layout:

```html
<tr>
    <td>Something...</td>
    <td colspan="7">Something...</td>
    <td>Something...</td>
</tr>
```

If every `<td>` has a colspan attribute, all of them get overwritten by the column count.

```html
<tr>
    <td colspan="2">Something...</td>
    <td colspan="3">Something...</td>
    <td colspan="2">Something...</td>
</tr>
```

To:

```html
<tr>
    <td colspan="7">Something...</td>
    <td colspan="7">Something...</td>
    <td colspan="7">Something...</td>
</tr>
```

### DIFF
--- a/js/api/api.row.details.js
+++ b/js/api/api.row.details.js
@@ -180,7 +180,12 @@ var __details_events = function ( settings )
 				row = data[i];
 
 				if ( row._details ) {
-					row._details.children('td[colspan]').attr('colspan', visible );
+					row._details.each(function () {
+						var el = $(this);
+						if (el.children('td').length == 1) {
+							el.children('td[colspan]').attr('colspan', visible);
+						}
+					});
 				}
 			}
 		} );
--- a/js/api/api.row.details.js
+++ b/js/api/api.row.details.js
@@ -166,6 +166,25 @@ var __details_events = function ( settings )
 			} );
 		} );
 
+		// Column visibility change - update the colspan
+		api.on( colvisEvent, function ( e, ctx, idx, vis ) {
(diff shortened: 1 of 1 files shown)


AllanJard:
Thanks for the PR. Unfortunately I can't pull it in, since that function is required for cases where a row template isn't being used. However, I do see the issue and I'll look at getting a fix for that.

murilo-plantae:
Thanks for the quick response Allan, do you think the next patch will include a fix for it ?

AllanJard:
I hope so. I'd look at it just now, but we've got a power cut here, so I'm making do with a phone atm...

murilo-plantae:
Do you think something like this fix the issue without breaking the feature when there has no row template?

```js
if ( row._details ) {
    row._details.each(function () {
	var el = $(this);
	if (el.children('td').length == 1)
            el.children('td[colspan]').attr('colspan', visible);
	});
}
```

AllanJard:
That is more or less what I had in mind, yes.

murilo-plantae:
I reverted the initial commit and commited the suggested code.

murilo-plantae:
Yes.

PR #250, opened 2023-11-16
Add dynamic defaults

manusfreedom:
Useful for retrieving default values from Ajax or for defining certain parameters using a function) :
```js
$.fn.dataTable.defaults = function(){
    myDefault = $.extend({}, $.fn.dataTable.defaults, { "serverSide": true });
    myDefault.stateSave = getFromAjax($(this).attr("id"), "stateSave");
    return myDefault;
}
```

### DIFF
--- a/js/core/core.constructor.js
+++ b/js/core/core.constructor.js
@@ -2,7 +2,7 @@
 var i=0, iLen, j, jLen, k, kLen;
 var sId = this.getAttribute( 'id' );
 var bInitHandedOff = false;
-var defaults = DataTable.defaults;
+var defaults = typeof DataTable.defaultsFunction === 'function' ? DataTable.defaultsFunction.call(this) : DataTable.defaults;
 var $this = $(this);
 
 


AllanJard:
Thanks for suggesting this. I'm not sure about it to be honest, but then that might just be because I'm not to sure where this would be useful over just setting the defaults in a common function. Are you able to give an example?

Thanks!

manusfreedom:
We mainly use Datatables for 50 different tables and store personal parameters in our database using an API, but each table has its own specific parameters (and some can be global and overloaded) such as stateSave. Each table has its own JS, so there's no need to put common managed (but personal) parameters on each JS.
We give users the choice of using a default saveState value, but this can be overridden in each table and linked to the user rather than the session/browser.

AllanJard:
Just reviewing some old issues - sorry this has been sitting for a while. I think this would cause a whole bunch of issues for backwards compatibility for libraries and extensions that expect `defaults` to be static. As such, I'm afraid this isn't something I'm going to pull in.

PR #258, opened 2024-02-16
Fix issue where an empty data set continues to display "Loading..." message

TheNephalim:
Hello,

I ran into an issue when I returned an empty record set from using server-side processing (SSP).  In these instances, the table continued to display "Loading..." instead of the empty message "No data available in table."  

I stepped through the code and it appeared the issue lie with the _emptyRow method located in the /js/core/core.draw.js source file.  

When the data table initially loads, the value of iDraw is 0 and the number of records for the table should be 0, so we display the loading message.  On the next pass through _emptyRow, iDraw should be 1 (or greater) and the number of records should be 0, so we should display the empty table message.

I'm sorry if I messed this up.  I'm relatively new to making pull requests.

### DIFF
--- a/js/core/core.draw.js
+++ b/js/core/core.draw.js
@@ -494,7 +494,7 @@ function _emptyRow ( settings ) {
 	var zero = oLang.sZeroRecords;
 	var dataSrc = _fnDataSource( settings );
 
-    if ( settings.iDraw  < 1 && settings.fnRecordsTotal() === 0 && (dataSrc === 'ajax' || dataSrc === 'ssp') )
+    if ( settings.iDraw  < 1 && (dataSrc === 'ajax' || dataSrc === 'ssp') )
 	{
 		zero = oLang.sLoadingRecords;
 	}
--- a/js/core/core.draw.js
+++ b/js/core/core.draw.js
@@ -494,7 +494,7 @@ function _emptyRow ( settings ) {
 	var zero = oLang.sZeroRecords;
 	var dataSrc = _fnDataSource( settings );
 
-	if ( settings.iDraw <= 1 && (dataSrc === 'ajax' || dataSrc === 'ssp') )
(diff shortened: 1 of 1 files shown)


AllanJard:
Hi - thanks for the PR!

Could you show me the response from the server when this happens? Is it completely empty (which would be valid JSON and should throw a JSON error) or is it an empty array or something else?

I should also note that the response from the server, when using server side processing (which I assume you are here?) should never have a draw value of 0. It will be 1 or higher.

TheNephalim:
Sure thing!  Here is the request:

{
  "requestVerificationToken": "1509A48661555C391712783B0AABECD6A8D83D5",
  "chgNum": "",
  "chgRev": "",
  "advKeyWrd": "bbb",
  "class": "Blah78",
  "chgType": "",
  "srchStat": "",
  "draw": 1,
  "maxRows": 25,
  "startAt": 1,
  "order": [
      {
          "column": 0,
          "dir": "asc",
          "name": ""
      }
  ]
}

And here is the response:

{
  "recordsFiltered": 0,
  "success": true,
  "recordsTotal": 0,
  "draw": 1,
  "data": [],
  "content": "Successfully retrieved change details"
}

As you can see, the response is valid JSON and the data is an empty array which, to the best of my knowledge, is valid, empty data per the requirements of the data table.  

I should note that the application behaves as expected up to 1.13.8.  When I install version 2.0.0,  I get the described problem.

AllanJard:
Yes, I would expect that to work, and am really surprised it doesn't. Apologies for that error. I'll try to recreate it shortly (on phone atm).

TheNephalim:
> Yes, I would expect that to work, and am really surprised it doesn't. Apologies for that error. I'll try to recreate it shortly (on phone atm).

You don't have to apologize.  I love this library and just want to help.  BTW, you are very responsive and I wasn't expecting a response so quickly.  

I edited my response above to add a few more details, but I don't know whether they really add anything germane to the discussion.

AllanJard:
Excellent stuff - thank you! It will be in 2.0.1 which should drop in the next few days.

PR #277, opened 2024-05-02
Performance: Do not apply sort when there is only one match

xPaw:
This shaves off ~100ms when calling `rows(.., {order: 'index'})` on a table with 10k rows because each row goes through selector stuff.

EDIT: Saves over a second when not using `order` option (so it uses current).

Before:
image

After:
image

This may be showing that there's some performance issue in iterating over `rows()` in general as it goes through all the selector and sorting shenanings for every row.


AllanJard:
Sounds good to me! Are you happy contributing this under the MIT license?

xPaw:
Yes.

AllanJard:
Thank you!

PR #280, opened 2024-05-07
Selector row indexes: should return same range when SSP regardless of…

jybleau:
… page

Update api.selectors.js:
Fixes responsive toggle icon only showing on page 1 and not on other pages.

Example: https://live.datatables.net/rufivaga/7/edit
In this example, the toggle in column 1 is only visible on page 1, but should also be visible on other pages.
This PR fixes this issue.


AllanJard:
Thanks for the PR. I'll need to have a think about if this is the correct solution. It might be next week before I get a chance to do a proper dive.

AllanJard:
Yes, I agree - I think this is the correct solution - nice one!

Are you happy for this to be contributed and included under the DataTables MIT license?

PR #285, opened 2024-07-02
Bootstrap3 col-xs support

Bamuel:
In Bootstrap3,

If you do col-sm-6
it will not work properly and need to specify col-xs-6

Before:
image

After:
image


Bamuel:
An option to manage the layout class would be awesome feature to resolve this issue.

Bamuel:
Thanks for implementing a solution 👍🏻 
https://github.com/DataTables/DataTablesSrc/releases/tag/2.1.0

PR #297, opened 2024-10-21
make scrollbody focusable

Lexachoc:
If there is a scrollable table, it would be nice to focus on the scrollbody so that we can use the keyboard to scroll.

This PR adds the tab index to the scrollbody according to the value:

https://github.com/DataTables/DataTablesSrc/blob/daa37b2f01e288acbfbff1e019cd3b9f44aeec7c/js/model/model.defaults.js#L491

Result:
recording

Note: It seems to be redundant if [KeyTable Plugin](https://datatables.net/extensions/keytable/) is used, as keyboard navigation is already available for cells.
Example: https://datatables.net/extensions/keytable/examples/initialisation/scrolling.html

### DIFF
--- a/js/core/core.scrolling.js
+++ b/js/core/core.scrolling.js
@@ -139,6 +139,8 @@ function _fnFeatureHtmlTable ( settings )
 		$(scrollBody).css('height', scrollY);
 	}
 
+	$(scrollBody).attr('tabindex', settings.iTabIndex)
+
 	settings.nScrollHead = scrollHead;
 	settings.nScrollBody = scrollBody;
 	settings.nScrollFoot = scrollFoot;


AllanJard:
Thanks for the suggestion! I rather like this - it is however the first time I've encountered this request, so I'm not going to immediately pull it in, but let's see if there is any other feedback from others who would like this.

pczarn2:
It looks like this example already has the focus on the tbody during tabbing: https://datatables.net/extensions/responsive/examples/styling/scrolling.html
However, the behavior changes once I click to sort by one of the columns.

pczarn2:
Here is a comment from our QA/Accessibility team member. 

Hmm, not 100% sure if I understand the question and situation, though I know enough to make one comment.

As long as just a keyboard can be used to navigate/scroll through a table where only some of it visible at a time, that seems fine. 

If someone using just a keyboard cannot see part of the data table, then that’s where the issue comes up.

Giving the scroll body keyboard focus does introduce an unpleasant interaction with screen readers, though. NVDA starts to read the contents of the ENTIRE table while JAWS still stuck and reads little/nothing, which is confusing for the user.

Lexachoc:
@pczarn2 Thank you! I completely overlooked the usecase of screen reader. (Sorry I have no experience with it). So it would be better not to focus the scroll body, but focus to the cells instead (as in the examples that have focusable cells). If someone want to do that, then they can manually use the render function ([example](https://datatables.net/examples/basic_init/data_rendering.html)) to do that.

This PR should then not be merged and developers should manually ensure the accessibility.

pczarn2:
I think so too. The first example you gave (https://datatables.net/extensions/keytable/examples/initialisation/simple.html) seems to be best use for this particular case.

PR #298, opened 2024-10-22
Accessibility fixes for ordering icons

pczarn2:
Some screen readers would read the content of the up/down arrows when tabbed to headings. Replaced the actual up/down arrow with css drawings to fix this issue.


AllanJard:
That is what the `/ ""` in the `content` is for - so screen readers wouldn't read it. What screen reader is it you are using that doesn't support that?

AllanJard:
No worries - thanks for checking it!

PR #300, opened 2024-10-27
fix typo in the doc for columns().types()

Lexachoc:
https://datatables.net/reference/api/columns().type()
should be
https://datatables.net/reference/api/columns().types()


AllanJard:
Doh - thank you for finding, reporting and fixing this!

PR #302, opened 2024-10-31
fix: date render helpers wrt. setting locales in Luxon

### LINKED ISSUE #301: Datetime render helpers are broken when using Luxon and an explicit locale argument
This is about the [build-in render helpers](https://datatables.net/manual/data/renderers#Built-in-helpers) provided by DataTables. Specifically about the date/datetime/time helpers when using them with an explicit locale argument and the Luxon backend.

Example: `DataTable.render.date("D", "de")` should return the timestamp as a "short date" in German format (i.e. "day.month.year") but it returns it in the (default) American format ("month/day/year") instead.

The helper implementation for Luxon ([source](https://github.com/DataTables/DataTablesSrc/blob/824b6febdbf26e7e60c08ef6c702f4226b6b85a4/js/ext/ext.helpers.js#L64)) calls `dt.setLocale(locale)` but then returns the original `dt` probably assuming that it's state has been mutated, but AFAIK that is not how Luxon works. Luxon uses an immutable data strategy and always returns a modified copy instead of mutating. So the line in question should be `dt = dt.setLocale(locale)`.

penenkel:
fixes #301

Luxon works with immutable objects, so we have to override the variable with its modified copy.

### DIFF
--- a/js/ext/ext.helpers.js
+++ b/js/ext/ext.helpers.js
@@ -61,7 +61,7 @@ function __mldObj (d, format, locale) {
 			return null;
 		}
 
-		dt.setLocale(locale);
+		dt = dt.setLocale(locale);
 	}
 	else if (! format) {
 		// No format given, must be ISO


AllanJard:
Excellent point - thanks for the fix!

PR #304, opened 2024-11-05
update doc for `dt-error`

Lexachoc:


Lexachoc:
BTW, it would be nice to have an example on how to listen to all error events from all tables (without using the `id`).

From https://datatables.net/reference/event/dt-error
> This event will bubble up the document, so you can add a listener for dt-error.dt to the body to capture all error events triggered by DataTables.

AllanJard:
Thank you! Good fix in the docs.

Example:

```js
$(document).on('dt-error.dt', function () {
  // ...
});
```

should do it.

PR #312, opened 2025-01-15
Bug fixes in the server side datatables example, specifically in spp.class.php

4lva:
The add bindings function contains an error that causes the binds to never be added to the array passed as a parameter, causing a SQL error when executing the statements.

In addition, in the complex function, the global bindings are not added to some statements.


4lva:
Yes, I agree to it being added under the MIT License.

AllanJard:
Great - thanks for the contribution!

PR #317, opened 2025-02-11
Add 'unfilteredKeys' option to exclude specific keys from global search

JFHarvey-Qc:
## Summary

This PR introduces a new option, `unfilteredKeys`, which allows developers to exclude specific columns from the global search in DataTables.

## Usage Example

```javascript
$('#ClientList').DataTable({
    data: dataSet,
    columns: [
        { data: 'id', visible: false },
        { data: 'client', title: 'Client Name' },
        { data: 'description', title: 'Description' }
    ],
    unfilteredKeys: ['id']  // Excludes the 'id' from global search
});


## Why This Change?
Currently, there is no native way to exclude certain columns from the global search without custom filters. This option simplifies the process by allowing developers to specify columns via their mData keys.

In the example, the key 'id' will not be used when filtering, but can still be used to identify the row.


## License Acknowledgment

I acknowledge that my contribution is made under the MIT license, as per the project's existing license.

### DIFF
--- a/examples/resources/demo.css
+++ b/examples/resources/demo.css
@@ -584,53 +584,53 @@ html.dark div.dt-tabs > div > ul > li:first-child {
 }
 html.dark code.option {
   color: #f7d6df;
-  background-color: #36373b;
-  border: 1px solid #2c2e32;
+  background-color: rgb(54.4, 54.7, 59.2);
+  border: 1px solid rgb(43.7, 45.85, 50.1);
 }
 html.dark code.path {
   color: #8ccb89;
-  background-color: #2c3633;
-  border: 1px solid #262d2e;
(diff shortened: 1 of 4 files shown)


AllanJard:
> Currently, there is no native way to exclude certain columns from the global search without custom filters.

This is exactly the  the [`columns.searchable` option](https://datatables.net/reference/option/columns.searchable) does.

See [this example](https://live.datatables.net/jujuzufi/1/edit) - the contents of the "Position" column aren't part of the search performed from the global search box.

I'm going to close this PR as I think the functionality already exists in DataTables, please please correct me if I've missed something.

JFHarvey-Qc:
You are quite right, I really missed that one.

Thank you for your time.

PR #324, opened 2025-03-02
Add `this` type to callback functions

xPaw:
Basically fixes the following:

```js
table.DataTable( {
	rowCallback( rowRaw, data, index )
	{
		const api = this.api(); // Property 'api' does not exist on type 'Config'.
	}
} );
```

I hope all the `Function*` on the bottom are used by Config, so change should be correct for all of them?


AllanJard:
That's a very good point - thank you.

PR #329, opened 2025-03-09
Typescript: Improve types of reduce() and reduceRight()

### LINKED ISSUE #327: Typescript: `reduce()` argument types are not correct
```
reduce(fn: ((current: number, value: any, index: number, dt: Api<any>) => number), initialValue?: any): any;
```

`current` is typed as `number`, but I think it needs to be the same type as the `initialValue` (so template it with `T`)

example code:
```js
apps = element.rows( '.app' ).nodes().reduce( function( coll, item )
{
	coll.push( item );
}, [] );
```

xPaw:
Fixes #327.

I copied these from lib.es5 types, and adapted them for datatables, so I'm hoping it'll be correct for all cases.

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -1148,9 +1148,9 @@ export interface Api<T=any> {
      * @param initialValue Value to use as the first argument of the first call to the fn callback.
      * @returns Result from the final call to the fn callback function.
      */
-    reduce(callbackfn: (current: T, value: T, index: number, dt: Api<any>) => T): T;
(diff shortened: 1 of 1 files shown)


AllanJard:
Looks good - many thanks for this. Are you happy for it to be included under the MIT license?

xPaw:
I'm fine with MIT in this and all my future PRs.

AllanJard:
Many thanks.

PR #330, opened 2025-03-10
Fix possible "Maximum call stack size exceeded."

xPaw:
I've had some mobile browsers reporting this error on some pages that had tables with upwards of 100k rows.

I've reproduced the issue on desktop chrome by making a table that had a million rows. Looking at this code, it doesn't even need to select all rows, just getting the first one is enough (maybe this selector could be improved even further).

I've created a jquery issue here: https://github.com/jquery/jquery/issues/5638

### DIFF
--- a/js/core/core.constructor.js
+++ b/js/core/core.constructor.js
@@ -239,7 +239,7 @@ _fnApplyColumnDefs( oSettings, oInit.aoColumnDefs, columnsInit, initHeaderLayout
 /* HTML5 attribute detection - build an mData object automatically if the
  * attributes are found
  */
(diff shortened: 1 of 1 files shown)


AllanJard:
I'm happy with that change - thanks for the PR! Are you happy for it to be included under the MIT license?

I ran into `.apply(...)` issues with a max stack size a while back and had [to put this in](https://github.com/DataTables/DataTablesSrc/blob/7a25d5a5d7b23d82afaf932d6106e53f222869a7/js/core/core.support.js#L318).

xPaw:
Yes I've fine with MIT.

PR #334, opened 2025-03-31
Fix OrderArray TS type

tiwe-ppi:
I used the [order API](https://datatables.net/reference/api/order()) and noticed that the result can contain empty strings and not just `'asc'` and `'desc'`. I don't know if `OrderIdx` and `OrderName` should be updated as well. I just fixed the part where I got a TS error. Let me know if I should update the two interfaces as well.

These changes can be included under the MIT license.


AllanJard:
That's a good point - thanks for that! An empty string means no order applied.

PR #336, opened 2025-04-03
Update Row Grouping example to use 2D arrays

kthorngren:
The [Row Grouping example](https://datatables.net/examples/advanced_init/row_grouping.html) uses 1D arrays with the `order()` API which breaks the use of `var currentOrder = table.order()[0];`.  This change is to use the 2D array data structure, that is consistent with the `order` option and clicking a column header to sort, in the event handler.  This is to fix the issue found in this [thread](https://datatables.net/forums/discussion/80788/the-example-for-row-grouping-will-not-sort-descending#latest).

This contribution follows the project's existing license (MIT).


AllanJard:
Nice one - thank you. It should be that the return from `order()` is consistent regardless of the input. That has actually been fixed for 2.3, but yup, I'm quite happy that the example reflects this.

PR #346, opened 2025-05-27
Add security fix for prototype pollution vulnerability

zaini:
This is a security patch release for DataTables 1.10.13 that addresses a prototype pollution vulnerability.

### Security Fix
- Added protection against prototype pollution attacks by preventing the use of `__proto__` in data paths
- The patch is a backport of the fix from commit [2632813d2ee433f7805f09f47a2afe6937c355ce](https://github.com/DataTables/DataTablesSrc/commit/2632813d2ee433f7805f09f47a2afe6937c355ce)

I would like this fix to be available as a patch on top of version `1.10.13` to avoid having to upgrade through all the other versions up to `1.10.22` to receive this fix.

---

Questions:

* What is your preference for the version name?

---

My contribution is offered under and will be made available under the project's existing license (MIT).

### DIFF
--- a/js/DataTables.js
+++ b/js/DataTables.js
@@ -1,11 +1,11 @@
-/*! DataTables 1.10.13
+/*! DataTables 1.10.13-sp1
  * ©2008-2016 SpryMedia Ltd - datatables.net/license
  */
 
 /**
  * @summary     DataTables
  * @description Paginate, search and order HTML tables
- * @version     1.10.13
+ * @version     1.10.13-sp1
  * @file        jquery.dataTables.js
  * @author      SpryMedia Ltd
  * @contact     www.datatables.net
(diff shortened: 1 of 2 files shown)


zaini:
Thanks for the prompt reply @AllanJard 

We've been using this package for years, and unfortunately reliant on some of the quirks of this set up so upgrading is tricky. For example, changes to the DOM structure and CSS classes and other behaviors make upgrading a tough option when we mainly just want to resolve this vuln.

In the case we cannot get a backport applied, we'll probably work off a fork of `1.10.13` until we are able to upgrade.

AllanJard:
Yes, sorry, I won't be merging patches back to unsupported versions at this time. If you need to use that specific version, rather than the bug fix versions for that minor release series, then yes, a fork would be the right thing to do.

PR #350, opened 2025-07-04
Fix: work in xhtml files

dmunozv04:
Hi, in  [xhtml, all tags must have closing tags](https://www.w3.org/TR/xhtml1/diffs.html#:~:text=Essentially%20this%20means%20that%20all%20elements%20must%20either%20have%20closing%20tags) and DataTables creates an element that doesn't have a closing tag, thus not loading correctly in pages distributed as html. 
Even though xhtml is not used frequently, trying to debug this issue in a website that uses it is painful (don't ask me how I know), and adding a closing tag wouldn't affect users that use DataTables on regular html5 pages.
If you need any additional info, let me know

### DIFF
--- a/js/core/core.sizing.js
+++ b/js/core/core.sizing.js
@@ -64,7 +64,7 @@ function _fnCalculateColumnWidths ( settings )
 		.removeAttr( 'id' );
 
 	// Clean up the table body
-	tmpTable.append('<tbody>')
+	tmpTable.append('<tbody/>')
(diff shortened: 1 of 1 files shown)


dmunozv04:
Thanks for taking a look so quickly!

Fair enough, I've gone ahead and updated it. 
happy for this to be included under the MIT license

dmunozv04:
> Yes, I'll aim to get that done by the end of the week.

Thanks!

PR #358, opened 2025-08-18
add Unix parsing on luxon library

xJuvi:
The momentjs library can parse various strings into a time object. This includes Unix timestamps in seconds (X) and milliseconds (x). The luxon library, on the other hand, cannot parse these two formats, only format them. With this PR, parsing for Unix timestamps is added manually to maintain the same range of functions.

### DIFF
--- a/js/ext/ext.helpers.js
+++ b/js/ext/ext.helpers.js
@@ -53,8 +53,8 @@ function __mldObj (d, format, locale) {
 		}
 	}
 	else if (__luxon) {
-		dt = format && typeof d === 'string'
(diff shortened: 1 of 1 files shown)


AllanJard:
Hi Hannes,

Have you tried asking in Luxon if they would consider adding support for unix epoch time stamps to `fromFormat`? I agree it is a small change, but my worry is that if I add support for this specific format since Luxon don't, then what other formats might Moment have and Luxon doesn't, and vice-versa. I'm afraid I'm still of a mind not to add this support at this time - it should just be a pass through to the date/time library and what formats are supported is up to them.

PR #373, opened 2026-01-30
Enable credentialled CORS request for language file

michael-dev:
In order to load the language file with a different domain with credentials enabled, `xhrFields: { withCredentials: true }` is needed in the `$.ajax` call. So pass xhrFields from oLanguage to `$.ajax`, like it is already possible when fetching the table data using ajax.

Example usage
```
DataTable({
    ajax: 'source.php',
    language: {
      url: 'https://example.org/js/datatable.de-DE.json',
      xhrFields: {
        withCredentials: true
      }
    },
    columns: [ ... ]
  });
```

Contribution under MIT license is acknowledged.

### DIFF
--- a/js/core/core.constructor.js
+++ b/js/core/core.constructor.js
@@ -382,6 +382,7 @@ if ( oLanguage.sUrl ) {
 	$.ajax( {
 		dataType: 'json',
 		url: oLanguage.sUrl,
+		xhrFields: oLanguage.xhrFields ?? {},
 		success: function ( json ) {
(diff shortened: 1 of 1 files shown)


AllanJard:
[This commit](https://github.com/DataTables/DataTablesSrc/commit/e715ec80ad4599dd1f3fca6378b96a39d2af0e0e) on the `3` branch adds the ability to use `language.ajax` as an object, extending properties for the Ajax request, just like is done for the main `ajax` option.

As such, I'm going to close this PR for now. Many thanks for sending it in and apologies for the delayed reply here.

Issue #210, opened 2022-06-07
Feature request: Columns should be searchable via the API, even if they were disabled during DataTable initialization


MisterVector:
Currently, attempting to search a column via the API when searching on that column has been disabled does not return anything. I'm posting this feature request because I think that some columns may not be appropriate to search on via the search field, however through one or more filters after a search, such a column ought to be searched on. I'll try to clarify with a proof of concept.

To play around with this POC yourself, I have provided the following gists:

eligibleEmployees.html: https://gist.github.com/MisterVector/6cbbb74b74d5af7556472ec8681a5b64
eligibleEmployees.js: https://gist.github.com/MisterVector/d8b7a6c3f7b02bfce2d268c86a720ea8

The POC starts off with this table:

![image](https://user-images.githubusercontent.com/437096/172299720-58d9d75a-1de5-4da4-8c27-18d380ab1c61.png)

Scenario 1 below illustrates my point that some columns aren't appropriate to search on:

![image](https://user-images.githubusercontent.com/437096/172300055-873eb0c1-69c8-4006-a541-e72c6e80a29a.png)

Scenario 2 shows the "Show only employees eligible for promotion" option checked, which shows only the employees eligible for promotion:

![image](https://user-images.githubusercontent.com/437096/172308228-778c0b2b-7f64-4ed8-9378-a8cd0ca83777.png)

With the "Disable third column for search?" item checked, scenario 3 shows that performing the same search will not show any results, as that column has been disabled:

![image](https://user-images.githubusercontent.com/437096/172307478-58ed7806-b36d-4c2c-b92d-44a5301f6cce.png)

For scenario 4, both checkboxes have been checked, which disables the third column for search and performs a search on that column using the API, which turns up no results:

![image](https://user-images.githubusercontent.com/437096/172308510-9eca62fb-782a-4857-ba84-0da0f247dec4.png)

I'm suggesting that scenario 4 should show the intended results, regardless if the searching capabilities of that column are disabled or not.

This probably makes more sense if the column was hidden. You would disable it at initialization to make sure the hidden column is not selected when searching but then perform a search on that column using the API.

AllanJard:
Yes, rather than just being a boolean option, it might be nice to allow it to be `searchable: 'api'`.

It isn't something I'm going to be adding soon though I'm afraid. This is the first request I can recall for this and it isn't something I've required myself yet. Will leave open for future thought.

MisterVector:
[...]

PR #377, opened 2026-03-12
Add the ability to configure a column to be searchable only by the API. Addresses #210

MisterVector:
This PR adds the ability to configure a column to be searchable only by the API. There may be a situation such as when a table contains a certain column for display purposes only, however there is a separate filter button or checkbox that is used to perform a search on that column. An example would be a list of employees at a company, a subset of which are eligible for promotion. Their eligibility is displayed in a column with the words "Yes" or "No." When using the search feature to search the table, if the first or last name of someone contains the letters "y", "e", "s", "n" or "o", we may not want the eligibility column to be used in the search. However, if we would like to quickly display the eligible employees using a button or checkbox for such purposes, we would need the column to be searchable by the API.

You can find a proof-of-concept [here](https://files.codespeak.org/projects/dt_pr/eligibleEmployees.htm).

This PR adds a new value for the `searchable` column property in the DataTable settings object. For a table with ID `myTable` with 3 columns, if you would like the second column to be searchable by the API only, you would need to do something like the following when initializing a datatable:

```javascript
new DataTable('#myTable', {
    columnDefs: [{ searchable: 'api', targets: 1 }]
});
```

A new filter cache has been added for column data that is searchable by the search feature as well as the API. This is because when a search is performed by the API, all searchable column data is available to search. When a search is performed with the search feature, column data available only to the API cannot be searched. Therefore, two filter caches need to be used.

The following code using a ternary operator was done for consistency sake although there doesn't appear to be a situation where a specific column is searched using the search feature. I'll let you decide if such code should stay or if I should change it to use `row.aFilterData[ column ]` exclusively

image

I acknowledge that my contribution will be licensed under the MIT license.

### DIFF
--- a/docs/option/columns.searchable.xml
+++ b/docs/option/columns.searchable.xml
@@ -4,11 +4,11 @@
 	<summary>Enable or disable search on the data in this column</summary>
 	<since>1.10</since>
 
-	<type type="boolean" />
+	<type type="boolean|string" />
 	<default value="true" />
 
 	<description>
-		Using this parameter, you can define if DataTables should include this column in the filterable data in the table. You may want to use this option to disable search on generated columns such as 'Edit' and 'Delete' buttons for example.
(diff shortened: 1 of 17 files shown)


AllanJard:
I've just merged changes into the `3` branch to add the sub-set column search ability. I'm going to be working on getting a beta of v3 out soon :-).

MisterVector:
Thank you very much! I am glad that there has finally been a resolution to this!

PR #389, opened 2026-05-31
Fix row details state restore: unescape escaped colons in row IDs

MatrixNeoKozak:
Fixed a bug in state restoration of child rows where colons in row IDs were not being unescaped after being escaped during state saving. The original code only escaped colons on save but did not unescape on load, causing the state to fail to restore child rows for IDs containing colons (e.g., 'a:b'). Added the reverse operation in detailsStateLoad to unescape the colon before passing to rows().


AllanJard:
I might be being daft here, but is the new map function just undoing the escaping that the previous one is adding? Are you able to give me a link to a test case that shows the issue so I can dig into it as well please?

PR #390, opened 2026-06-17
Run non-jQuery tests without jQuery

ottenhoff:
Thanks for this project. We use DataTables quite a bit throughout the open-source LMS https://github.com/sakaiproject/sakai. We are hoping to move all of our various jQuery-backed table plugins to DataTables 3.0.0 without jQuery. 

For this PR, the nonjQuery tests were loading jQuery via dt.libs, so they weren't exercising the no-jQuery code path. This commit removes jQuery from those suites and updates the shared test loader to work without jquery.

### DIFF
--- a/test/html-loader-lib.js
+++ b/test/html-loader-lib.js
@@ -50,11 +50,13 @@
 		// CHeck if everything is loaded
 		if (queue.length === 0) {
 			// Add a jQuery function to trigger native event handlers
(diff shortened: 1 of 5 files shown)


AllanJard:
Good call - thank you. Are you happy with me merging this under the MIT license?

ottenhoff:
> Good call - thank you. Are you happy with me merging this under the MIT license?

Yes!

PR #391, opened 2026-06-17
Fix vanilla delegated blur handling

ottenhoff:
DataTables maps non-bubbling events to bubbling equivalents for delegated handlers. `blur` was mapped to `blurout`, but the browser event is `focusout` (MDN: https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event).

I also added a no-jQuery regression test for delegated `blur`.

### DIFF
--- a/js/dom/events.ts
+++ b/js/dom/events.ts
@@ -136,7 +136,7 @@ function parseEventName(original: string | null) {
 		isFocus = true;
 	}
 	else if (name === 'blur') {
-		name = 'blurout';
+		name = 'focusout';
 		isFocus = true;
 	}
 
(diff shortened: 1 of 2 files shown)


AllanJard:
Awesome - thank you.

Happy with MIT license again (just need to check with each PR, rather than having a CLA)?

ottenhoff:
All my contributions to DataTables, now and in the future, may be used under the MIT license.

Thanks!

AllanJard:
Thank you.

I'll try to remember and link back to this PR, but knowing me I'll forgot and ask in each PR if it isn't already explicitly stated :-).

PR #396, opened 2026-06-19
Cancel Ajax when beforeSend returns false

### LINKED ISSUE #393: js/util/ajax.ts sends after abort
if beforeSend returns false, it calls `xhr.abort()` but still calls `xhr.send(...)`. That should cancel and return without sending.

ottenhoff:
Fixes #393.

This contribution is offered under the DataTables project's MIT license.


AllanJard:
Nice catch - thank you :-)

PR #400, opened 2026-07-28
Update some missing/incorrectly named JSDoc @param items

miken32:
Noticed the two in `dom/index.ts` but then figured I'd just fix them all!

### DIFF
--- a/js/api/Api.ts
+++ b/js/api/Api.ts
@@ -515,7 +515,7 @@ function getPrototypeNames(name: string) {
  * Each of the input parameter types will be converted to a DataTables settings
  * object where possible.
  *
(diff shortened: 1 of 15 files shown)


AllanJard:
Brilliant! Thank you! Out of interest, how did you find all of these? Does your IDE highlight them?

Are you happy for these changed to be brought in and distributed under the MIT license?

miken32:
Yes, I just inspected the files in PhpStorm and looked for that particular error. And please feel free to distribute my contributions under any license you see fit.

AllanJard:
Smashing - thank you.

There is probably a similar plugin for VSCode (which I tend to use). I'll take a look :-)

PR #401, opened 2026-07-29
Fix headless loading via. commonJS

### LINKED ISSUE #385: Headless loading of "datatables.net-bs5" fails if global.window defined.
Hi there, if you'd rather this be a forum post then apologies, but hopefully this is a bug rather than me being silly.

I'm writing unit tests for some of my DT customisations, to do this I'm using JSDOM to feed them enough of a window to do their work. Currently (DT 2.3.8) running the following script in nodeJS falls over:

```javascript
import DataTable from 'datatables.net-bs5';
import { JSDOM } from 'jsdom';

// NB: Storing the window object globally, so my extensions find it at the usual place
global.window = (new JSDOM('<!DOCTYPE html><html><body></body></html>')).window;

const dt = DataTable(global.window);
/**         $.extend( _ext, {
          ^

TypeError: $.extend is not a function
    at /srv/node_modules/datatables.net/js/dataTables.js:998:4
    at /srv/node_modules/datatables.net/js/dataTables.js:36:21
    at Object.<anonymous> (/srv/node_modules/datatables.net/js/dataTables.js:43:2)
    at Module._compile (node:internal/modules/cjs/loader:1529:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1613:10)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1096:12)
    at Module.require (node:internal/modules/cjs/loader:1298:19)
    at require (node:internal/modules/helpers:182:18)
    at cjsRequires (/srv/node_modules/datatables.net-bs5/js/dataTables.bootstrap5.js:17:5)

Node.js v20.19.2
*/

// Do a thing to show dt loaded successfully
console.log(dt.ext.oStdClasses.container);
```

Over in ``dataTables.bootstrap5.js`` we have:

```javascript
                var jq = require('jquery');
                var cjsRequires = function (root, $) {
                        if ( ! $.fn.dataTable ) {
                                require('datatables.net')(root, $);
                        }
                };

                if (typeof window === 'undefined') {
                        module.exports = function (root, $) {
                                if ( ! root ) {
                                        // CommonJS environments without a window global must pass a
                                        // root. This will give an error otherwise
                                        root = window;
                                }

                                if ( ! $ ) {
                                        $ = jq( root );
                                }

                                cjsRequires( root, $ );
                                return factory( $, root, root.document );
                        };
                }
                else {
                        cjsRequires( window, jq );
                        module.exports = factory( jq, window, window.document );
                }
```

When creating a DT in a headless environment, ``dataTables.bootstrap5.js`` only calls ``cjsRequires`` once we're instantiating an object. This then triggers the factory machinery in ``dataTables.js``. However, by that point my fake window is already in place and it tries to configure itself for a browser environment, which is where the rot sets in.

I think the fix is something along the lines of:

```diff
--- /srv/node_modules/datatables.net-bs5/js/dataTables.bootstrap5.js.o        2026-04-28 09:00:41.662090967 +0000
+++ /srv//node_modules/datatables.net-bs5/js/dataTables.bootstrap5.js  2026-04-28 09:00:46.622137064 +0000
@@ -12,9 +12,10 @@
        else if ( typeof exports === 'object' ) {
                // CommonJS
                var jq = require('jquery');
+               var baseDt = require('datatables.net');
                var cjsRequires = function (root, $) {
                        if ( ! $.fn.dataTable ) {
-                               require('datatables.net')(root, $);
+                               baseDt(root, $);
                        }
                };

```

...so the base factory is called at the same time (and thus in the same environment) as the bs5 one.

It's not quite as simple as that mind, as [this code is autogenerated by wrapper.js](https://github.com/DataTables/DataTablesSrc/blob/6e7650e79da74bde9779f59a3cbb24d3384851f4/build/wrapper.js#L118-L140). But could form a dict of the required modules to then call. If you'd like me to have a go at this then can do.

I can work around this by adding a DT to JSDOM before I set ``global.window``. Then because we don't re-attach DT the second call works. Feels like a bit of a kludge though.

lentinj:
I've been tinkering with this on and off for a bit with DT 3 as suggested. The following fix #385 to the point that a script like this can run in nodeJS:

```javascript
import { JSDOM } from 'jsdom';

// Vanilla DT import or bootstrap DT import (both should work now)
import DataTable from './DataTablesSrc/built/DataTables/js/dataTables.js';
//import DataTable from './DataTablesSrc/built/DataTables/js/dataTables.bootstrap5.js';

console.log("--------------------- Creating JSDOM window");
const jsd_window = (new JSDOM('<!DOCTYPE html><html><body></body></html>')).window;
global.window = jsd_window; // NB: Creating a global window object to potentially confuse the factory function

console.log("--------------------- Creating DataTable");
const dt = DataTable(jsd_window);
console.log(dt.ext.oStdClasses.container);
```

Using EJS however is a bit of a non-starter:

* Extension-less imports (including the ones in ``dataTables.bootstrap5.mjs``) will load the ``.js`` version without ``type: "module"`` in ``package.json``. How safe this is to add I'm not sure.
* A by-product of the CommonJS mechanism is they make the ``window`` & ``document`` objects available to other code. In EJS-land, this doesn't happen, there's a long string of code (e.g. the line in the second commit) that runs on import that assumes that window or document are defined.

There doesn't seem to be an obvious solution for the latter. Going round and doing ``globalThis.window ? ...`` seems sisyphean, and wouldn't be the right window object anyway. Maybe a separate headless wrapper would work, but I've not looked into how to construct one of those yet.

And thanks again for dataTables!

My contribution is offered under and will be made available under the project's existing license (MIT).

### DIFF
--- a/build/wrapper.js
+++ b/build/wrapper.js
@@ -165,15 +165,19 @@ ${exportProps}
 function umd(script, deps, exp, filename) {
 	let amdLoad = [];
 	let commonjs = [];
+	let commonjsrequires = "";
 	let requiresDT = false;
 
+	commonjsmodules = "{" + deps.map((dep) => {
+		return `'${dep}': require('${dep}')`;
+	}).join(",") + "}";
 	for (let dep of deps) {
 		amdLoad.push(`'${dep}'`);
 
 		if (nameFromDependency(dep) === 'DataTable') {
 			commonjs.push(`
 			if (! root.DataTable) {
-				require('${dep}')(root);
+				root.DataTable = cjsModules['${dep}'](root);
 			}
 `);
 
@@ -183,8 +187,9 @@ function umd(script, deps, exp, filename) {
 			let name = nameFromDependency(dep);
 
 			commonjs.push(`
-			if (! window.DataTable.${name}) {
-				require('${dep}')(root);
+			baseModule['${dep}'] = require('${dep}');
+			if (! root.DataTable.${name}) {
+				root.DataTable.${name} = cjsModules['${dep}'](root);
 			}
 `);
 		}
@@ -221,6 +226,7 @@ function umd(script, deps, exp, filename) {
 	}
 	else if (typeof exports === 'object') {
 		// CommonJS
+		var cjsModules = ${commonjsmodules};
 		var cjsRequires = function (root) {${commonjs.join('')}		};
 
 		if (typeof window === 'undefined') {
--- a/js/util/ajax.ts
+++ b/js/util/ajax.ts
@@ -44,7 +44,7 @@ const defaults = {
 	contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
 	headers: {},
 	traditional: false,
-	url: location.href
+	url: window.location.href
 } as AjaxOptions;
 
 /**


AllanJard:
Interesting - many thanks for this. Is it specifically CommonJS that you want to use, or is that just a workaround as headless doesn't work with ES Modules (ESM, which is what I presume you mean by EJS)? Looking at the code above, finding some way to have the ESM loaded work for a headless environment looks more desirable (so you can use the styling packages).

AllanJard:
Out of interest, what is the use case for using DataTables headless - server-side rendering of the table?

AllanJard:
I've just been looking a bit more into using DataTables in a windowless environment, and I do have it setup to allow the `window` (and thus also `document`) property with the `DataTable.use()` static function, so you should be able to do something like:

```js
import DataTable from 'datatables.net/js/dataTables.mjs';
import { JSDOM } from 'jsdom';

const jsd_window = new JSDOM(
	`<!DOCTYPE html>
	<html>
		<body>
			<table id="myTable"></table>
		</body>
	</html>`
).window;

DataTable.use(jsd_window);

const dt = new DataTable('#myTable');
```

However, that very definitely doesn't work. I've got a handful of "pre-init" setup functions that use window properties, such as checking the locale number format.

I'd really like to get this addressed so it does work in native ESM environments (it should also then work for CommonJS since it is loader independent), however while checking this I've also realised that my package.json doesn't use `exports` so Node.js doesn't see the ESM file, which is another change to be made. 

What I think I'll do is set this as a 3.1 target. The changes are a bit too extensive to get into 3.0.1 (or any 3.0 patch). That shouldn't be _too_ far away. I want to get StateRestore and SearchPanes ported to v3, then consider this. Apologies it isn't going to be a super fast fix.

lentinj:
> Is it specifically CommonJS that you want to use, or is that just a workaround as headless doesn't work with ES Modules (ESM, which is what I presume you mean by EJS)?

I'm not fussed, anything that works (and yes, sorry for getting the names wrong!)

> what is the use case for using DataTables headless 

Unit-testing modules containing custom column renderer functions that build on ``DataTable.render``. I don't want a full DT instance, just enough that I can get at the utility classes and then exercise my own renderer functions.

The module is still browser code, so I'd like to keep doing ``import DataTable from 'datatables.net-bs5'`` at the top and then the tests do whatever they need to make that work. Currently that includes using CommonJS because of nodeJS' ``type: "module"`` shenanigans.

Getting a factory back in some cases is something that caused head-scratching, but sounds like this is already something you're picking apart. An aside, [jQuery deprecated theirs with 4.0.0 and have a separate factory import](https://github.com/jquery/jquery/commit/46f6e3da796ee9d28c7c1428793b72d66bcbb0b7).

> What I think I'll do is set this as a 3.1 target

Yeah, that's more than fine. I did start looking at detangling the pre-init uses of window/document, but it clearly needed more structural thought.

I've conflated 2 things a bit here, quick fixes needed to improve CommonJS support in nodeJS and the issue of using ESM in nodeJS. If you'd like me to do the issue gardening to separate the latter into it's own issue then can do. If you'd rather not merge the pull request that's also fine, I think I can work around everything at this stage.

Thanks again for your help!

AllanJard:
Hi - many thanks for your reply and additional thoughts on this.

Yes, I think it probably wouldn't be right to pull in this PR at the moment, knowing that a full fix for both issues (which go hand in hand) will be coming in 3.1.

I'll leave this issue open for the moment though, as window-less is currently not going to work. I'll update when its done :)

incadoi:
Hi - we also have this issue in our project, the tests are failing using the latest version. Curious to know if you plan to release 3.1 soon. Thanks!

AllanJard:
My current expectation is that it is around 3 months away. I need to finish rewriting StateRestore and SeachPanes for DataTables 3, there are a couple of other things I want to do related to the web-site and DataTables usage, and then hopefully I can focus on DT3.1. There are a bunch of changes that will be needed for that as well, so unfortunately it will take time I'm afraid.

You could perhaps lock your version to DT2 for the moment? It had better support for a windowless environment, although it was a bit quirky (which is why it changed for DT3).

incadoi:
ok, thanks for the reply!

AllanJard:
I've just committed a change to a new [3.1 branch](https://github.com/DataTables/DataTablesSrc/tree/dt3.1) that helps address the issues here. There is a proviso though - DataTables uses a lot of browser DOM provided objects such as `Node`, `Option`, etc. I could go through the code updating every reference, and will do if there is significant demand for this feature, but at the moment what you need to do is copy the objects to the Node global space - e.g.:

```
import DataTable from 'datatables.net/js/dataTables.mjs';
import { JSDOM } from 'jsdom';

const dom = new JSDOM(
	`<!DOCTYPE html>
	<html>
		<body>
			<table id="myTable"></table>
		</body>
	</html>`,
	{url: 'http://localhost/'}
);
const window = dom.window;

// Copy JSDOM window properties to Node.js global context
Object.getOwnPropertyNames(window)
	.filter((key) => !(key in globalThis))
	.forEach((key) => {
		globalThis[key] = window[key];
	});

DataTable.use(window);

const dt = new DataTable('#myTable');

console.log(dt.table().container().innerHTML);
```

Note also that the `url` property needs to be specified, otherwise `localStorage` throws an error while doing this. It would be possible to filter to just the objects that DataTables uses, and while that might not be too bad for the core, the extensions will use a bunch as well.

If you get a chance to try it out, I'd welcome that. Going to aim to rotate to 3.1 release fairly quickly - I'd like this change and a change in the npm packages for better native ESM support to be the core parts.

